PR #15, opened 2023-12-15
Increase test coverage

lukasjhan:
## Purpose

Measure test coverage and increase test coverage


pensivej (changes requested):
It would be nice if you separate measure/displaying coverage and increasing coverage

lukasjhan:
> It would be nice if you separate measure/displaying coverage and increasing coverage

I remove measure/displaying coverage here and create new PR at #22

pensivej (approved):

Issue #30, opened 2023-12-18
Add key binding example


lukasjhan:
## Overview

Add key binding example in example project.

- Issue with key binding
- verify sd jwt with key binding

PR #31, opened 2023-12-18
[#30]Add key binding example

lukasjhan:
Issue: #30 

## Purpose

Add key binding example. I already add description in wiki


lukasjhan:
Currently kb example verifiy failed due to #36 bug.

pensivej (approved):

Issue #45, opened 2024-01-26
Support 07 version of standard


lukasjhan:
* support new version: 07 standard 
* revise verify, validate return type (boolean -> { header, payload }
* support custom signer, verifier function

PR #46, opened 2024-01-26
Support 07 version of standard and new features for v2.0

lukasjhan:
Issue: #45

## Purpose

Add features for v2.0:
- Support 07 version of standard
- revise verify, validate return type (boolean -> { header, payload }
- support custom signer, verifier function

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -98,7 +98,7 @@ export class SDJwtInstance {
       header,
       payload: {
         ...packedClaims,
-        sd_alg: options?.hash_alg ?? SDJwtInstance.DEFAULT_HASH_ALG,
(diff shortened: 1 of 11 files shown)


src/index.ts
```diff
@@ -95,7 +98,7 @@ export class SDJwtInstance {
       header,
       payload: {
         ...packedClaims,
-        _sd_alg: options?.hash_alg ?? SDJwtInstance.DEFAULT_HASH_ALG,
+        sd_alg: options?.hash_alg ?? SDJwtInstance.DEFAULT_HASH_ALG,
```
lukasjhan, line 101: This is change for standard version 07

lukasjhan:
_sd_alg is not changed. _sd_hash is changed to sd_hash in key binding JWT.
my mistakes, I'll fix it

zustkeeper (approved):

PR #57, opened 2024-02-07
fix: the bytes of the output of the hash function must be base64url-encoded.

y12studio:


lukasjhan:
Hi @y12studio.

> The bytes of the output of the hash function MUST be base64url-encoded
> https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-07.html#section-5.2.3

Like you pointed out, There is a bug processing hash of disclosure.
But I think we should modify the digest function to return base64url in crypto.ts. Because user's hasher function can be used and it might not return a hex string.

User's hasher function is passed through,

* [index.ts] issue function pass hasher function to pack function
* [sdjwt.ts] pack function use it to disclosure.digest and createDecoy

It would be good if you could share your thoughts please. :)

lukasjhan:
Maybe, `User's hasher function must return a base64url string` should be added to the docs later.

y12studio:
@lukasjhan I encountered issues during `pnpm test` while trying to modify the digest function to directly return a base64url-encoded string. Despite my efforts, I couldn't resolve these errors. As a temporary solution, I've implemented a workaround that allows the tests to pass successfully.

To ensure long-term maintainability and code quality, I propose merging this PR first. This will allow you to address the digest function rewrite without compromising test integrity.

lukasjhan:
That would be a good idea. :)

lukasjhan (approved):

lukasjhan:
>@lukasjhan I encountered issues during pnpm test while trying to modify the digest function to directly return a base64url-encoded string. Despite my efforts, I couldn't resolve these errors.

I'll check on this :)

PR #62, opened 2024-02-13
feat: use tsup as a bundler

cre8:
Instead of tsc, it using https://tsup.egoist.dev/ to generate commonjs and esm


lukasjhan:
I'll test the bundle and the browser testing :)

lukasjhan:
@cre8 I got a quick question. 

When I build the package with tsup, tsup concat all the code into one js file and one d.ts file. (or one mjs and one d.mts, I guess this is ESM). Can I still get the benefits of tree shaking even if I only have one file?

cre8:
@lukasjhan Tree shaking works great with ESM, since it can analyse the exports and remove usued code before bundling. Therefore putting everything will not have a negative impact on the build size, see: https://esbuild.github.io/api/#tree-shaking
It also has no negative side for debugging since we are adding the .d.ts files.

tsub however has a "legacy" mode that will create a separate folder for esm and cjs and compile every file for it's own.

lukasjhan:
> @lukasjhan Tree shaking works great with ESM, since it can analyse the exports and remove unused code before bundling. Therefore putting everything will not have a negative impact on the build size, see: https://esbuild.github.io/api/#tree-shaking It also has no negative side for debugging since we are adding the .d.ts files.
> 
> tsub however has a "legacy" mode that will create a separate folder for esm and cjs and compile every file for its own: https://tsup.egoist.dev/#bundle-formats

Good :) I'm currently resolved conflict and fixing tests on the top of your branch.
You can check my progress here: https://github.com/lukasjhan/sd-jwt-js/tree/feat/esm%2Bcjs-fix

lukasjhan:
This is really strange in here https://github.com/lukasjhan/sd-jwt-js/tree/feat/esm%2Bcjs-fix. 
I'm still using node crypto in test files, but test:broswer succeed.

cre8:
@lukasjhan I found no way to use window.subtle function inside the jsdom, it seems that is not integrated. Right now I am building a dummy web app to test it there directly to verify to correctness in the browser

cre8:
@lukasjhan works nice in the browser with esm and the test are passing, see: https://github.com/cre8/sd-jwt-test

lukasjhan (approved):
@cre8 Wow I tried your test repo, It tests with the broswer.
Maybe jsdom just mock the DOM, not the whole broswer. We can try to integrate broswer later :)

I think PR is good

cre8:
Since we are not using window.crypto in the actual code base, it should be fine that it is not covered by jsdom and we can still achieve 100%.
In case we include it, we can mock it for the test (or look for a parallel test that is supporting browser, but for now jsdom should be fine)

Issue #44, opened 2024-01-04
Add CD for npm publish


lukasjhan:
## Overview

OWF provide npm token for our repo. you can check it in discord channel.
We'll apply same way with https://github.com/openwallet-foundation/agent-framework-javascript/blob/main/.github/workflows/continuous-deployment.yml

But, I think it should be triggered by release

cre8:
One good approach I saw:
- the code on main will always be published under the tag next
- versions will be managed by creating a new release
- release documents are auto generated by semantic release commits lik `fix: ...`

lukasjhan:
That approach seems good. Currently I deploy it manually but OWF provide us a npm token, so next time, let's use it to publish automatically

I think triggering the release to the npm when we make a new release in github is better. :)
What do you think? @cre8

cre8:
Before choosing a way, I would look at the used framework that should be used to manage a monorepo when we want to split it up.

E.g. Lerna has a lot of functions already for versioning and i would prefer to reuse these common approaches so we can focus to work on the lib and not the "managing stuff" around it.

lukasjhan:
I'll make a PR for integration with Lerna.

lukasjhan:
@berendsliedrecht @cre8 

Based #68, now we can discuss the package names. I believe @openwalletfounction/* will be used since there is a token provided by OWF. 

I made name in this format: sd-jwt-{sub package name}. so e.g. sd-jwt-type, sd-jwt-core.
I would appreciate it if you could share your thoughts please :)

cre8:
the prefix with openwalletfoundation is fine for me, since owf is already taken.

also the approach with the general prefix `sd-jwt` is fine for me.

To make the versioning easier I would suggest to sync the version numbers over the packages. In this case when we release something, we have it like:
```
Version 1.1.0
sd-jwt-core:
- fixed bug foo bar
```

lukasjhan:
Version sync is great. If not, there will be a compatible issue between each versions

lukasjhan:
I guess there is no package named @owf/*.
How can I check owf is used or not? @cre8

cre8:
Any chances we can reuse the @sd-jwt that is used right now?

Other typescript projects from the owf like credo also use their own prefix.

berendsliedrecht:
We can use the `@sd-jwt`, yes! Also, installing, or importing, multiple packages with the `@openetfoundation` can get quite annoying as it is fairly long :).

[...]

PR #71, opened 2024-02-20
WIP: Feat/versioning

cre8:
Update of this PR:

vitest usage: vitest was configured for workspace usage. This will generate a coverage report for all files (open todos: publish result file to codecoverage for better inspection. We need to set a token as a secret).

adding lint: to improve code quality lint was added. But for not it is outcomment during the ci since it would not pass ;)

optimized ci: I used veramos one as a reference. It will build and test the code. It will also publish the code to NPM when it is on one of the specific branches
- if its on main, it will publish the latest version and also with a new version number based on the tags. It will also generate a release
- if its on next, it will publish the latest version under the next tag, but will not give it a new version number
- similar to next, but it will publish it with the unstable tag.


lukasjhan (dismissed):
Looks good to me :)
Is it ready to merge?

cre8:
@lukasjhan not yet

cre8:
@lukasjhan since I am not getting the results I want with lerna, I will try to migrate to NX which seems to be more up to date how to handle monorepos.

lukasjhan:
@cre8 The direction of change seems to be good, but what part is not working well?

cre8:
> @cre8 The direction of change seems to be good, but what part is not working well?

The continuous deployment is not working. Lerna has the call to generate the changelog and the updated version in the package jsons. But I found no good way to push these changes to the GitHub branch inside the GitHub action. I also found no documentation on the Lerna website.

Three options:
- we are doing the release manually in the short run
- we are checking how other projects using Lerna in their cicd have solved it
- we switch to other tools like nx where are plugins to publish

I already tested the switch to nx and it was done in 10 minutes.

I will evaluate all three options today so we merge this PR successful

lukasjhan:
I'll put sd-jwt-vc PR after this :)

cre8:
@lukasjhan I updated this entry so it can be reviewed again by you :)

cre8:
I am not sure if the publish step will work at this point because I cannot check if we have all the required variables and secrets set

lukasjhan:
Okay :) I'm reviewing right now.
The DCO check is failed. Do you want to bypass it?

cre8:
I will fix the the DCO check by resigning in an hour.

cre8:
@lukasjhan done, an hour is relative :D

CHANGELOG.md
```diff
@@ -0,0 +1,11 @@
+# Change Log
+
+All notable changes to this project will be documented in this file.
+See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
+
+## [2.0.3](https://github.com/openwallet-foundation-labs/sd-jwt-js/compare/v2.0.2...v2.0.3) (2024-02-20)
```
lukasjhan, line 6: I assume that this changelog generated automatically, is it right?
It might be needed to fix manually as we discuss in #44.  To 0.x

lukasjhan (approved):

CHANGELOG.md
cre8: I think we should remove the generated one from this PR. As mentioned above, I don't know what's the best way to generate the changelog that should be part of the github repo. It should be relatively easy to use the release mechanism of github

lukasjhan: @cre8 I think we might need to reset all the tags in this repo when we deploy to @sd-jwt/*

lukasjhan: I guess release mechanism of github is easiest way :D.

PR #77, opened 2024-02-22
Rename packages into @sd-jwt

lukasjhan:
Changes

- change name of packages
- remove changelog.md
- fix package.json workspace(?)


packages/decode/package.json
```diff
@@ -34,11 +34,11 @@
   },
   "license": "Apache-2.0",
   "devDependencies": {
-    "@hopae/sd-jwt-node-crypto": "2.0.3"
+    "@sd-jwt/crypto-nodejs": "workspace:*"
   },
   "dependencies": {
-    "@hopae/sd-jwt-type": "2.0.3",
-    "@hopae/sd-jwt-util": "2.0.3"
+    "@sd-jwt/types": "workspace:*",
```
lukasjhan, line 40: @cre8 This part has actual version. So I changed it to workspace:* again.
I think lerna publish set the version here. Was this intentional?

berendsliedrecht: Yeah `workspace:*` should be swapped by lerna for the release. You can check the setup at https://github.com/berendsliedrecht/sd-jwt-ts as we used the exact same setup there.

PR #88, opened 2024-02-23
test: exclude coverage dir

lukasjhan:
I exclude the coverage dir from the testing.


cre8:
I don't think that is required since we tell vitest to only include the package/* folder. Therefore we do not need to exclude coverage since it's not a sub directory from packages.

I also could not find the coverage folder on the generated coverage report. And since the coverage rate seems to be the same at this PR and the e2e PR, I would say it has no effect.

@lukasjhan am I missing something?

PR #89, opened 2024-02-23
fix: ci publish action

cre8:
closes: #44 

Updates the used variables so the action runs.

@lukasjhan 
we need to define what is the "primary" branch. Veramo has this logic:
- next is defined as the "default" branch where everything is merged. A commit to this branch will make a release for "next".
- main is the branch holding the latest release. The difference between next and main is that not for every PR a new version is generated, but Authors are able to decide when it's time to create a new release on github (and to publish a new release version on npmjs). This will also set the :latest version on npmjs.
- unstable: allows to push a version to npmjs for testing. It is not required to be first pushed to unstable, then to next and then to main. I am not quite sure if we need this stage. Maybe we can remove it in the future

This approach allows to have a tested version approach (because when we saw some bug on the next release, we can overwrite it in an easy way). On the other hand it's still the classic approach for the user because he can use latest or no version number during the installation to make sure he gets the latest released version.

### DIFF
--- a/.github/workflows/build-test-publish-on-push-cached.yaml
+++ b/.github/workflows/build-test-publish-on-push-cached.yaml
@@ -117,18 +117,18 @@ jobs:
       - test
       - lint
     env:
-      NPM_TOKEN: ${{secrets.NPM_TOKEN}}
-      NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
-      GH_TOKEN: ${{secrets.GH_TOKEN}}
-      GITHUB_TOKEN: ${{secrets.GH_TOKEN}}
-      GH_USER: ${{secrets.GH_USER}}
-      GH_EMAIL: ${{secrets.GH_EMAIL}}
+      NPM_TOKEN: ${{secrets.NPM_TOKEN }}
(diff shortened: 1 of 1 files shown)


lukasjhan (dismissed):

lukasjhan:
@cre8 How about managing with main, release, unstable?

lukasjhan:
> @cre8 How about managing with main, release, unstable?

I thought changing the default branch would be tricky, so I thought I'd just make the default branch "main".
But I think your point is reasonable. next means next lastest candidate.
Let's use next branch as you proposed 🙂

lukasjhan:
Let's merge it after default branch is changed.

cre8:
Keeping this PR open until #91 is done to avoid conflicts

cre8:
@lukasjhan I added the settings file that should set the default branch to next. we are also able to define other rules in this file so #91 can be closed with this PR

lukasjhan (approved):
Good. Can we merge it? @cre8

cre8:
@lukasjhan I think so

cre8:
@lukasjhan I had to open another PR to add the `--no-private` step

PR #93, opened 2024-02-23
fix: add publish config

cre8:
adding `publishConfig` with access:public

the no-private param only is to exclude private packages from being processed.


lukasjhan (approved):

lukasjhan:
Good :)

PR #106, opened 2024-02-25
feat: es crypto features to nodejs and browser

lukasjhan:
I implement 

- Ed25519 and ES256 signer and verifier for crypto-nodejs
- ES256 signer and verifier for crypto-browser (not tested)


lukasjhan:
hmm.. I have no idea how to test browser crypto subtle module. Any ideas?

cre8:
> hmm.. I have no idea how to test browser crypto subtle module. Any ideas?

It depends on the test framework and I read that vitest with jsdom does not support the crypto apis... I tested it manually by running a crypress test with angular because then it is using a chrome instance in the background

lukasjhan:
@cre8 Yeah, we definitely use some framework that use headless chrome(or other browser)

How about trying this? https://playwright.dev/

cre8:
> @cre8 Yeah, we definitely use some framework that use headless chrome(or other browser)
> 
> How about trying this? https://playwright.dev/

I can take a look at this

lukasjhan:
> > @cre8 Yeah, we definitely use some framework that use headless chrome(or other browser)
> > How about trying this? https://playwright.dev/
> 
> I can take a look at this

Thank you :)

lukasjhan:
I want to support Ed25519 and ES256 as JWK but,

browser - doesn't seem to support Ed25519 natively.
node - bad jwk support

I turned this PR to draft and keep working on it.

lukasjhan:
In nodejs I used [webcrypto](https://nodejs.org/api/webcrypto.html#web-crypto-api)

but it is only support node >= 20. I tested in node 18 it failed.

lukasjhan:
@cre8 Looks like test are working well ;)

cre8:
@lukasjhan I think we need to support ES256 since it's the only algorithm supported by both platforms. Of course we can implement dozens of others, but in the first place we should have one algorithm people can use in both environments to sign and verify credentials.

Are you sure with the support? According to the docs webcrypto is supported since v15.

lukasjhan:
@cre8 I applied your matrix check :) 
I remove Ed25519 in nodejs-crypto package. Like you said, At first let's supporting ES256

cre8 (approved):
LGTM

PR #109, opened 2024-02-25
chore: add missing tests for browser

cre8:
Adding missing tests for the brower.

@lukasjhan since we have a coverage of 100% for browser and I am calling the window.crypto api, it seems that vitest is able to interact with this.

I haven't tested your PR with this yet


lukasjhan (approved):
Good Let's merge it. I'll test my pr after this.

PR #111, opened 2024-02-25
Add matrix for testing

cre8:
adding the matrix testing approach allows us to test it for different versions

### DIFF
--- a/.github/workflows/build-test-publish-on-push-cached.yaml
+++ b/.github/workflows/build-test-publish-on-push-cached.yaml
@@ -43,7 +43,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: ['18.x', '20.x', 'stable']
+        node-version: ['16.x', '18.x', '20.x']
     steps:
       - uses: pnpm/action-setup@v3
         with:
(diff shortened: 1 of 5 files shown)


berendsliedrecht (approved):

cre8:
@lukasjhan you are correct, it seems that version 18 is and lower is not supported. At this point we have to decide:
- should the crypto-node package include native support with a higher version number
- should it be able to use them with lower support but with the help of external libraries

I haven't looked if we can implement the whole crypto-node stuff without the subtle. Code reusage is not possible then, but on the other hand the amount of functions is small that need to be implemented

cre8:
So the good news is that the node code is working fine in nodejs v16 18 and 20 :)

But I needed to outcomment the browser test since these tests seem not to be performed as a browser test, but a node test and therefore the tests failed since node subtle is not supported with all requirements.

We should try to find a solution for the browser tests to get them up and running again before merging this

cre8:
Found a good solution:
according to this issue `jsdom` is not supporting webcrypto, therefore the tests fail for node 16 and 18. For some reasons it's possible to call it in node 20, so I added a condition that the browser tests are only run in the matrix test when it's executed with at least 20.
Since we are only using the coverage of v20, we get the full coverage support for this. Since we do not have any version specific conditions like "only run this code when you are using v16, otherwise use this".

So we can also make sure the whole code base is running on node 16. I didn't check lower since from my point of view support 3 LTS versions should be enough.

lukasjhan:
3 TLS looks good :)

Issue #112, opened 2024-02-26
Add protection rule for next branch


cre8:
Like the main branch someone should not be possible to directly push to the next branch since this bypasses the review process

PR #113, opened 2024-02-26
Set protection rules for main and next

cre8:
Closes: #112


lukasjhan (approved):
Looks good!

PR #116, opened 2024-02-27
doc: add online debugger

lukasjhan:
I added [online debugger](https://sdjwt.co) in https://sdjwt.js.org


berendsliedrecht (approved):

cre8:
@lukasjhan I would also mention it in the root readme since this is a good tool for validation

lukasjhan:
> @lukasjhan I would also mention it in the root readme since this is a good tool for validation

Oh That is good idea! I added it to root README :) @cre8

cre8 (approved):
LGTM

PR #118, opened 2024-02-27
doc: fix installation command

lukasjhan:
I fixed the installation command in https://sdjwt.js.org.
It was out-dated.


cre8:
Depending on the conversation in the other PR we need the users to use `@sd-jwt/sd-jwt-vc` instead of the core package.

After we decided how to merge the other PR, I would then merge this one

lukasjhan:
> Depending on the conversation in the other PR we need the users to use `@sd-jwt/sd-jwt-vc` instead of the core package. And then we also need to update it maybe on other places.
> 
> After we decided how to merge the other PR, I would then merge this one

Okay That sounds good. :)

PR #119, opened 2024-02-27
doc: update github page when next branch updated

lukasjhan:
I update the github page deploy action target branch
Our page will be updated with PR to next branch.

### DIFF
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -4,7 +4,7 @@ name: Deploy static content to Pages
 on:
   # Runs on pushes targeting the default branch
   push:
-    branches: ["main"]
+    branches: ['next']
 
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:


cre8:
The github page should be focused on the changes that are deployed to "main" aka latest. Otherwise people are confused when the run `npm i @sd-jwt/core` and don't see a feature that is listed on the github page,

Possible solution: generate a github page for the next and main branch. This could be more difficult, so to keep it simple I prefer to stay with the "next" / current approach.

lukasjhan:
> The github page should be focused on the changes that are deployed to "main" aka latest. Otherwise people are confused when the run `npm i @sd-jwt/core` and don't see a feature that is listed on the github page,
> 
> Possible solution: generate a github page for the next and main branch. This could be more difficult, so to keep it simple I prefer to stay with the "next" / current approach.

That is a good point. People can be confused when they check the working in progress stuff. And managing two version of page is too heavy. Let's just keep the current approach. I'll close the PR

PR #120, opened 2024-02-28
feat: fixing exclude _sd_alg when no disclosure

lukasjhan:
I always set `_sd_alg` in payload even if there is no disclosure. But I think it's better not to include when there is nothing.
I guess it's more complaint.

> When used, this claim MUST appear at the top level of the SD-JWT payload.

> The payload MAY contain the _sd_alg key described in [Section 5.1.1](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-07.html#hash_function_claim).

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -100,7 +100,7 @@ export class SDJwtInstance {
       header,
       payload: {
         ...packedClaims,
-        _sd_alg: hashAlg,
+        _sd_alg: disclosureFrame ? hashAlg : undefined,
       },
     });
     await this.SignJwt(jwt);


cre8:
I reached out to the authors since the paragraph is confusing.

"If the _sd_alg claim is not present at the top level, a default value of sha-256 MUST be used."

I understand that the _sd_alg needs only be included in case I want to use a other algorithm than _sd_alg

lukasjhan:
> I reached out to the authors since the paragraph is confusing.
> 
> "If the _sd_alg claim is not present at the top level, a default value of sha-256 MUST be used."
> 
> I understand that the _sd_alg needs only be included in case I want to use a other algorithm than _sd_alg

FYI: @berendsliedrecht 's implementation doesn't have `_sd_alg` when there is no disclosure.

hmm... I thought that is for backward compatibility. I think excluding `_sd_alg` when using sha-256,  It can be good. It's more compact

cre8:
> > I reached out to the authors since the paragraph is confusing.
> > "If the _sd_alg claim is not present at the top level, a default value of sha-256 MUST be used."
> > I understand that the _sd_alg needs only be included in case I want to use a other algorithm than _sd_alg
> 
> FYI: @berendsliedrecht 's implementation doesn't have `_sd_alg` when there is no disclosure.
> 
> hmm... I thought that is for backward compatibility. I think excluding `_sd_alg` when using sha-256, It can be good. It's more compact

Your change is not violating the spec. And the small amount of chars in the JWT will have no negative effect. So from my point we can merge this and move on :)

cre8 (approved):

berendsliedrecht:
I did not include it, because when I started to implement on a lower draft version it basically stated that when no selectively disclosable items are included, it should be treated as a normal JWT. However, I think the point more is for verification as @cre8 mentioned, if it is not included assume sha2-256, but it does not always have to be set by default to sha2-256. If that makes sense.

PR #122, opened 2024-02-28
Make sdjwt instance non abstract

cre8:
As discussed in #114 the sdjwtinstance should be non abstract to allow people to use it on its own.

I included the `validateReservedFields` function that will return nothing in the sdjwtinstance, but can be implemented in extended classes like `sdjwtvcinstance`. Doing so, we don't have to override the while isse function

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -32,7 +32,8 @@ Ensure you have Node.js installed as a prerequisite.
 
 ### Usage
 
-This library can not be used on it's own, it is a dependency for other implementations like `@sd-jwt/sd-jwt-vc`.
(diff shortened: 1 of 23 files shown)


packages/core/README.md
```diff
@@ -32,7 +32,8 @@ Ensure you have Node.js installed as a prerequisite.
 
 ### Usage
 
-This library can not be used on it's own, it is a dependency for other implementations like `@sd-jwt/sd-jwt-vc`.
+The library can be used to create sd-jwt based credentials. To be compliant with the  `sd-jwt-vc` standard, you can use the `@sd-jwt/sd-jwt-vc` that is implementing this spec.
+If you want to use the pure sd-jwt class or implement your own sd-jwt credential approach, you can use this library.
```
lukasjhan, line 36: Good. I'll change the package name in #118

lukasjhan (approved):
It looks good to me :)

PR #127, opened 2024-02-29
Implement kbverify function

### LINKED ISSUE #125: feat: kbVerify improvment
right now the `kbVerify` uses the same input as Verify (payload and signature of the jwt). However the payload does not include a reference to the holder public key, therefore we are not able to verify the signature

cre8:
closes #125 

Adds signature verification to the kbVerify function. The whole payload of the actual token is passed to the kbverify function, it is not needed in the constructor of the KBJWT since this would break the extension from JWT.

The fetching of the key has to be implemented by the user. The payload has an optional field [cnf](https://datatracker.ietf.org/doc/html/rfc7800#section-3.2). The user is free to get the key from another field like `subject`, `id` etc. and is also free to parse the value - no limitation if there is a did url, jwk or url pointing to a hosted public key.<

Since we need to pass the payload to the verify function, the verify function in KBJWT will not get overwritten, but there will be a new function. It looks better to do it this way than adding an optional field to the verify function in JWT that has no meaning just to be compliant with the other call. If someone know a more elegant way I am open for changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
     "lint": "biome lint ./packages",
     "format": "biome format . --write",
     "test": "vitest run --coverage.enabled=true --coverage.include=packages/*",
+    "test:watch": "vitest",
     "clean": "lerna clean -y",
     "publish:latest": "lerna publish --no-private --conventional-commits --include-merged-tags --create-release github --yes --dist-tag latest",
     "publish:next": "lerna publish --no-private --conventional-prerelease --force-publish --canary --no-git-tag-version --include-merged-tags --preid next --pre-dist-tag next --yes",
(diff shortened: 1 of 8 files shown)


cre8:
Missing tests will be added in the next hours.

lukasjhan:
It looks good to me. :)

lukasjhan:
@cre8 Can we merge it?

lukasjhan:
> @lukasjhan I would love to implement the missing tests first. If you have the time please make a PR to my branch with the missing tests :)

Okay I'll make a test for the new verifyKB function :)

lukasjhan:
I added it in https://github.com/cre8/sd-jwt-js/pull/5

PR #129, opened 2024-03-03
feat: create kb jwt when present all

lukasjhan:
In present method, when `presentationKeys` is undefined, then present all disclosures. but in this case kbjwt is not created.
I fixed it to create kbjwt.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -146,7 +146,6 @@ export class SDJwtInstance<ExtendedPayload extends SdJwtPayload> {
       kb?: KBOptions;
     },
   ): Promise<SDJWTCompact> {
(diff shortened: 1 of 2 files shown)


packages/core/src/index.ts
```diff
@@ -146,17 +146,19 @@ export class SDJwtInstance<ExtendedPayload extends SdJwtPayload> {
       kb?: KBOptions;
     },
   ): Promise<SDJWTCompact> {
-    if (!presentationKeys) return encodedSDJwt;
```
lukasjhan, line 149: Due to the early return, kbjwt is not created.

cre8:
Good point, since the key binding has nothing to do with the selective disclosure, but with the holder binding.

cre8 (approved):

PR #130, opened 2024-03-03
feat: add jwk type

lukasjhan:
JsonWebKey type is in lib.dom.d.ts but some project might not want to include it tsconfig.json.
When I was trying to build credo-ts, It failed due to this. So I added it to our type file.

### DIFF
--- a/packages/core/src/test/index.spec.ts
+++ b/packages/core/src/test/index.spec.ts
@@ -1,9 +1,8 @@
 import { SDJwtInstance, SdJwtPayload } from '../index';
-import { Signer, Verifier } from '@sd-jwt/types';
+import { Signer, Verifier, KbVerifier, JwtPayload } from '@sd-jwt/types';
 import Crypto, { KeyLike } from 'node:crypto';
 import { describe, expect, test } from 'vitest';
 import { digest, generateSalt } from '@sd-jwt/crypto-nodejs';
-import { KbVerifier, JwtPayload } from '@sd-jwt/types';
 import { importJWK, exportJWK, JWK } from 'jose';
 
 export const createSignerVerifier = () => {
--- a/packages/types/src/type.ts
+++ b/packages/types/src/type.ts
@@ -136,14 +136,14 @@ type Frame<Payload> = Payload extends Array<infer U>
     ? Record<number, Frame<U>> & SD<Payload> & DECOY
     : SD<Payload> & DECOY
   : Payload extends Record<string, unknown>
-  ? NonNever<
-      {
-        [K in keyof Payload]?: Payload[K] extends object
-          ? Frame<Payload[K]>
-          : never;
-      } & SD<Payload> &
-        DECOY
-    >
-  : SD<Payload> & DECOY;
+    ? NonNever<
+        {
+          [K in keyof Payload]?: Payload[K] extends object
+            ? Frame<Payload[K]>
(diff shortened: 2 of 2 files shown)


cre8:
I would not include it as a type. as you mentioned it's already included in the jsdom type, together with all the other key types like rsa, ec, etc. So to keep our type files smaller and clean I would not include it.

cre8:
Right now we are not referencing `lib.dom.ts` in your configs, are we? The problem I had when I import `JsonWebKey` from e.g. `node:crypto`, then there are some type mismatching.

At this point I would prefer the way via lib.dom.ts since this is (as far as I know) natively supported in typescript.

lukasjhan:
> Right now we are not referencing `lib.dom.ts` in your configs, are we? The problem I had when I import `JsonWebKey` from e.g. `node:crypto`, then there are some type mismatching.
> 
> At this point I would prefer the way via lib.dom.ts since this is (as far as I know) natively supported in typescript.

That's good point :) I should find the way credo can use jsonwebkey that natively supported in typescript.

lukasjhan:
@cre8 I guess credo-ts wants us to include the type not natively.
https://github.com/openwallet-foundation/credo-ts/pull/1787#discussion_r1510675642

TimoGlastra:
@cre8 it seems a bit overkill to me to depend on lib dom just for JsonWebKey, and i also don't think it's needed to make the sd-jwt json web key type as complex as the lib dom one (i think only a few fields need to be defined and the rest can be typed as unknown).

Including lib dom can cause issues with global types in React Native or Node.JS, so it doesn't make sense to me to depend on browser types in an environment agnostic library. 

Not adding lib.dom will also make it easier for people to add sd-jwt to their project, as they don't need to add lib dom (which will add a lot of globals which are only available in the browser)

cre8:
> @cre8 it seems a bit overkill to me to depend on lib dom just for JsonWebKey, and i also don't think it's needed to make the sd-jwt json web key type as complex as the lib dom one (i think only a few fields need to be defined and the rest can be typed as unknown).
> 
> Including lib dom can cause issues with global types in React Native or Node.JS, so it doesn't make sense to me to depend on browser types in an environment agnostic library.
> 
> Not adding lib.dom will also make it easier for people to add sd-jwt to their project, as they don't need to add lib dom (which will add a lot of globals which are only available in the browser)

@TimoGlastra I will check the credo case, because when implementing the veramo plugin I did not need to include the dom lib. And our libs got compiled on node 16,18,20 with the latest typescript version.

Instead of defining the the jsonwebkey type in this lib, we could also pass the type as a parameter so every environment is able to use the required one.

lukasjhan:
I made a sample repo 🙂 https://github.com/lukasjhan/ts-lib-checking
After I was digging more, I found out that it is okay if skipLibCheck in tsconfig.json is true.

cre8:
> I made a sample repo 🙂 https://github.com/lukasjhan/ts-lib-checking After I was digging more, I found out that it is okay if skipLibCheck in tsconfig.json is true.

Ty Lukas :)

I will check the two options with its pros and cons:
- define jsonwebkey in this project
- pass it as type parameter

cre8:
@lukasjhan I removed the RSA key type, the `JsonWebKey` interface alone should be fine

cre8 (approved):

PR #131, opened 2024-03-04
feat: fix async handling in jwt verify

lukasjhan:
I found a bug while developing https://sdjwt.co

in jwt verify method, It doesn't wait until its resolved. verifier can be async function so we need await keyword.

### DIFF
--- a/packages/core/src/jwt.ts
+++ b/packages/core/src/jwt.ts
@@ -98,7 +98,7 @@ export class Jwt<
     const payload = Base64urlEncode(JSON.stringify(this.payload));
     const data = `${header}.${payload}`;
 
-    const verified = verifier(data, this.signature);
+    const verified = await verifier(data, this.signature);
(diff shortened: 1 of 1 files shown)


cre8 (approved):
Lgtm,

Have you also checked the code if we have other mistreated async calls?

And I think all defined functions that can be passed to the sd-jwt instance should be async, since we don't know if the user needs it for his implementations. Or we define some type like `Promise<string> | string`

lukasjhan:
> Lgtm,
> 
> Have you also checked the code if we have other mistreated async calls?
> 
> And I think all defined functions that can be passed to the sd-jwt instance should be async, since we don't know if the user needs it for his implementations. Or we define some type like `Promise<string> | string`

Yeah, I checked all code in core :) I guess now we're good.
Currently we treated passed functions as OrPromise<type> which is Promise<type> | type.

PR #133, opened 2024-03-05
feat: checking kb header value

lukasjhan:
I added checking logic of header of key binding JWT.

I referenced from standard:

> 8.3.  Verification by the Verifier
> 3. Ensure that a signing algorithm was used that was deemede public key for the Holder from the SD-JWT
>     secure for the application.  Refer to [RFC8725], Sections 3.1
>     and 3.2 for details.  The none algorithm MUST NOT be accepted
>  5.  Check that the typ of the Key Binding JWT is kb+jwt.

### DIFF
--- a/packages/core/src/kbjwt.ts
+++ b/packages/core/src/kbjwt.ts
@@ -9,22 +9,25 @@ export class KBJwt<
   // Checking the validity of the key binding jwt
   // the type unknown is not good, but we don't know at this point how to get the public key of the signer, this is defined in the kbVerifier
   public async verifyKB(values: { verifier: KbVerifier; payload: JwtPayload }) {
+    if (!this.header || !this.payload || !this.signature) {
+      throw new SDJWTException('Verify Error: Invalid JWT');
+    }
+
     if (
-      !this.header?.alg ||
+      !this.header.alg ||
+      this.header.alg !== 'none' ||
       !this.header.typ ||
-      !this.payload?.iat ||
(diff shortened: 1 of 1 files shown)


packages/core/src/kbjwt.ts
```diff
@@ -9,23 +15,26 @@ export class KBJwt<
   // Checking the validity of the key binding jwt
   // the type unknown is not good, but we don't know at this point how to get the public key of the signer, this is defined in the kbVerifier
   public async verifyKB(values: { verifier: KbVerifier; payload: JwtPayload }) {
+    if (!this.header || !this.payload || !this.signature) {
+      throw new SDJWTException('Verify Error: Invalid JWT');
+    }
+
     if (
-      !this.header?.alg ||
+      !this.header.alg ||
```
berendsliedrecht, line 23: Fine for now, but I think having an `assertHeaderClaim` method will help a lot here to make the code more readable and we can also return messages that tell the user on which property it failed.

like this: https://github.com/berendsliedrecht/sd-jwt-ts/blob/a1fb5fad0b938081e033233bae5569ae3b9fe8cb/packages/core/src/sdJwtVc/sdJwtVc.ts#L41

berendsliedrecht (approved):

packages/core/src/kbjwt.ts
lukasjhan: Oh, That's good idea.

PR #136, opened 2024-03-06
type: remove rsa type in jwk

lukasjhan:
I accidentally forgot to remove the rsa type.

### DIFF
--- a/packages/types/src/type.ts
+++ b/packages/types/src/type.ts
@@ -46,7 +46,6 @@ interface JsonWebKey {
   key_ops?: string[];
   kty?: string;
   n?: string;
-  oth?: RsaOtherPrimesInfo[];
   p?: string;
   q?: string;
   qi?: string;


zustkeeper (approved):

berendsliedrecht:
What was the reason for removing rsa here?

lukasjhan:
@berendsliedrecht We discussed this issue in #130. I thought @cre8 removed and approved it but It was not actually removed, So I opened this PR

berendsliedrecht:
Ahh okay, yeah that makes sense! If we fully want to delete it would should also get rid of the `n` and `e` fields. I don't think they are used by any other algo, but they are the main components for RSA keys.

lukasjhan:
> Ahh okay, yeah that makes sense! If we fully want to delete it would should also get rid of the `n` and `e` fields. I don't think they are used by any other algo, but they are the main components for RSA keys.

Oh I see, I'll prepare PR for that :)

cre8:
I think we could use the definition from the `dom.lib.ts` and just copy it. Would this solve all the edge cases?

cre8:
I would add the fields again. Just because some organisation say RSA usage is not safe anymore, it is still used out there ;)

With my comments on 130 I intended that we have just the `Jsonwebkey` interface, but inside it all attributes to cover different keys like in the lib.dom.ts

lukasjhan:
@cre8 @berendsliedrecht I opened a PR for restoring rsa field in JWK. #137

PR #137, opened 2024-03-06
feat: restore rsa type in JWK

lukasjhan:
I added the missing part(rsa) in JWK type

### DIFF
--- a/packages/types/src/type.ts
+++ b/packages/types/src/type.ts
@@ -34,6 +34,13 @@ export type KBOptions = {
   payload: Omit<kbPayload, 'sd_hash'>;
 };
 
+// This type declaration is from lib.dom.ts
+interface RsaOtherPrimesInfo {
+  d?: string;
+  r?: string;
+  t?: string;
+}
+
 interface JsonWebKey {
   alg?: string;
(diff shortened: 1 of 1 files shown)


cre8:
I don't think that's correct. The values for RSA have to be on the same level as the others, don't they?

lukasjhan:
> I don't think that's correct. The values for RSA have to be on the same level as the others, don't they?

But this is the implementation from typescript. 
https://github.com/microsoft/TypeScript/blob/main/src/lib/dom.generated.d.ts#L689
https://github.com/microsoft/TypeScript/blob/main/src/lib/dom.generated.d.ts#L1719

cre8:
> > I don't think that's correct. The values for RSA have to be on the same level as the others, don't they?
> 
> But this is the implementation from typescript. 
> https://github.com/microsoft/TypeScript/blob/main/src/lib/dom.generated.d.ts#L689
> https://github.com/microsoft/TypeScript/blob/main/src/lib/dom.generated.d.ts#L1719

You are correct, I looked at another spec. From this point of view the change is fine!

cre8 (approved):

PR #138, opened 2024-03-07
Fix/presentation transform

lukasjhan:
I changed the presentationKey input to presentation frame input.
If it's okay, then I'll change the examples :)

I also changed the type in type.ts

Feel free to share your thoughts.


berendsliedrecht (approved):

lukasjhan:
> @lukasjhan do we also need to update the docs because of this change? I don't mind if it's done in this PR or another one

Yeah, we need to updates our docs and examples. I'll do it right away after merge this :)

PR #147, opened 2024-03-08
chore(ci): merge from main into next

cre8:
this action will open a PR to next so a release that was generated on main will be synced with next


lukasjhan (approved):
Looks good to me :)

PR #151, opened 2024-03-09
feat: make _digest value public in disclosure

### LINKED ISSUE #150: Disclosure digest should be accessible without providing Hasher
For a holder, you receive the Disclosures with digests and right now can't easily access the digests of disclosures (private member). The only way to access a digest, is via calling `digest` which requires a HaserAndAlg as input
```typescript
  public async digest(hash: HasherAndAlg): Promise<string> {
    const { hasher, alg } = hash;
    if (!this._digest) {
      const hash = await hasher(this.encode(), alg);
      this._digest = Uint8ArrayToBase64Url(hash);
    }
    return this._digest;
  }
```

I do think it would be beneficial to make either the variable directly accessible (public) or change the signature of the function to have `hash` as an optional argument as it is only used when computing a new digest.

lukasjhan:
I made _digest value in disclosure as public. 

When implementing features in holder side, usually we decode the disclosures and make a logic. In this situation we normally have _digest value, not need to calculate. So delivering hasher to get _digest value is not necessary.

It closes #150

### DIFF
--- a/.github/workflows/build-test-publish-on-push-cached.yaml
+++ b/.github/workflows/build-test-publish-on-push-cached.yaml
@@ -107,12 +107,12 @@ jobs:
         with:
           node-version: 20
           cache: 'pnpm'
-      - name: Setup Biome
-        uses: biomejs/setup-biome@v2
-        with:
-          version: latest
+      # - name: Setup Biome
+      #   uses: biomejs/setup-biome@v2
+      #   with:
+      #     version: latest
       - name: Run Biome
-        run: biome ci .
+        run: npx biome ci .
 
   # Only run this job when the push is on main, next or unstable
   publish:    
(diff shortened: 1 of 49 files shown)


lukasjhan:
Hmm.. It's strange. biome success in local env

cre8:
> Hmm.. It's strange. biome success in local env

Same for me. Running biome ci . or pnpm run format or lint locally is fine....

cre8:
@lukasjhan found the error:
Biome release a new minor update, that added more rules. Locally, we are using version `1.5.0`, but when we used the github action, it is always going for latest, in our case `1.6.0`. So there were some new rules that causes the issue. I applied the changes, tests are passing.

cre8:
I referenced the biome call to the package json to make sure we got the same versions and therefore the same behavior local and in the ci.

cre8 (approved):

lukasjhan:
@cre8 Thank you :)

PR #153, opened 2024-03-11
Feat: select disclosure for present

lukasjhan:
I made a selectDisclosures function to select serialized Disclosures with PresentationFrame.
I didn't know there would be a use case like this, but PEX needs this function.

@cre8  I'm planning to integrate it in PEX right away, if it's merged after review, will I be able to distribute it as 0.4.1?

### DIFF
--- a/packages/present/src/present.ts
+++ b/packages/present/src/present.ts
@@ -177,6 +177,10 @@ export const selectDisclosures = <T extends Record<string, unknown>>(
   disclosures: SerializedDisclosure[],
   presentationFrame: PresentationFrame<T>,
 ) => {
+  if (disclosures.length === 0) {
+    return [];
+  }
+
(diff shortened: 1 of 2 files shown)


cre8:
@lukasjhan LGTM, can you add the missing tests so codecov is happy?

Releasing a patch is fine for me

lukasjhan:
> @lukasjhan LGTM, can you add the missing test so codecov is happy?
> 
> Releasing a patch is fine for me

Okay :)

lukasjhan:
@cre8 I found that I can't reproduce error in this implementation. I think we can't cover that 4 lines of throw exception

cre8:
@lukasjhan If we can not reproduce it, why are we checking it? :D

cre8 (approved):

lukasjhan:
Thank you. I never think of that :)

PR #164, opened 2024-03-12
doc: add npm download count in README

lukasjhan:
I added the npm download count in our README :)


cre8:
maybe also add the ci passing badge?

PR #174, opened 2024-03-12
feat: set token for deployment

cre8:
The release workflow will update the code by writing a new version number to the files. But since we have a protection rule on the main branch, we can not push directly. So the solution is to use a personal github token where the user is authorised to bypass the call.

In the future it would be better to use an app for this, see this discussion.
https://github.com/orgs/community/discussions/13836


lukasjhan (approved):

lukasjhan:
LGTM :)

PR #176, opened 2024-03-13
fix: incorrect condition on workflow step

charsleysa:
publish next step is incorrectly checking if the branch is `next` when the job only executes on the `main` branch causing publish next step to never execute


cre8 (approved):
LGTM, I forgot to update line 165

PR #179, opened 2024-03-15
fix: base64url convention

### LINKED ISSUE #178: Change the util/base64 function names to start with lowercase
By following the common javascript convention, I suggest all function names to be camelCase (or at least start with lowercase).
Especially when remarks can be misunderstood as a type instead of a function when imported somewhere else.

current

```javascript
import { Base64 } from 'js-base64';

export const Base64urlEncode = Base64.encodeURI;

export const Base64urlDecode = Base64.decode;

export const Uint8ArrayToBase64Url = (input: Uint8Array): string =>
  Base64.fromUint8Array(input, true);
```

to-be

```javascript
import { Base64 } from 'js-base64';

export const base64UrlEncode = Base64.encodeURI;

export const base64UrlDecode = Base64.decode;

export const uint8ArrayToBase64Url = (input: Uint8Array): string =>
  Base64.fromUint8Array(input, true);

```

lukasjhan:
I updated convention of base64url related functions.
This will be a breaking changes on @sd-jwt/utils. Maybe applied PEX but It's not merged yet, So It will be fine.

It will close #178

### DIFF
--- a/packages/core/src/decoy.ts
+++ b/packages/core/src/decoy.ts
@@ -1,5 +1,5 @@
 import type { HasherAndAlg, SaltGenerator } from '@sd-jwt/types';
-import { Uint8ArrayToBase64Url } from '@sd-jwt/utils';
+import { uint8ArrayToBase64Url } from '@sd-jwt/utils';
 
 // This function creates a decoy value that can be used to obscure SD JWT payload.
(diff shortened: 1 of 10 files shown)


cre8 (approved):
LGTM,
i would not frame it as breaking change, since we just changed the naming

PR #183, opened 2024-03-18
Fix/pass types

cre8:
- passes the type of claims to the present function. This makes sure in case the variable `presentationFrame` was initialised without a type.
- removes the object type from the transform function


lukasjhan (approved):
LGTM

PR #185, opened 2024-03-18
Feat/husky

cre8:
Adding husky to work to format the files before commiting them.


cre8:
@lukasjhan When I am using the commit via the command line like `git commit -m "fix: foo,....."` it works fine. It seems that vscode git commit ui is ignoring the pre commit, so it will create two commits (which is not a big deal since all commits will be squashed).

I used husky because it had the biggest community

lukasjhan (approved):
LGTM

berendsliedrecht:
Not the biggest fan of husky as the CI validates everything and husky just slows me down :'). No real issue, just find it a bit annoying and thats mainly why we disabled it within Credo.

lukasjhan:
> Not the biggest fan of husky as the CI validates everything and husky just slows me down :'). No real issue, just find it a bit annoying and thats mainly why we disabled it within Credo.

I've never used it before but, um.. Is it because husky check the codes every time I commit?

cre8:
@berendsliedrecht the problem with lerna publish is that it will update the package.json files to add the new version number, commit and push it. We can not stop in the middle to run the format command so it get's aligned with biome.

Or we go the other way around and change the biome rules so they fit with the changes lerna is performing.

lukasjhan:
I think it will be okay since the code base is not large like credo.

berendsliedrecht:
Yeah the size and biome performance should help a lot. As long as we don't have to wait for 50s+ for a commit it's all good :).

PR #208, opened 2024-04-18
Fix/drop node16

### LINKED ISSUE #207: Fix test CI for nodejs 16
Looks like new pnpm has version limit.

check here: https://github.com/openwallet-foundation-labs/sd-jwt-js/actions/runs/8707141840/job/23959522487#step:5:72

Maybe we need to use older pnpm to run the old version node tests

cre8:
closes #207 

since pnpm v9 dropped support and version 16 is out of the supported range, we should remove it


lukasjhan (approved):
LGTM :)

PR #212, opened 2024-04-24
fix: revert Bump vite from 5.2.9 to 5.2.10

lukasjhan:
I revert #209 since it failed publish with lerna.


cre8:
@lukasjhan ci is still failing, I will look at this

Issue #213, opened 2024-04-25
Bug: broken ci


cre8:
the publish task will fail, it seems the lock file is generated and lerna does not like uncommited changes.

I will try to find the reason why it's generated, "dirty fix" would be to ignore uncommited changes when lerna is running.

lukasjhan:
I think it's due to the new version of pnpm. When I use `pnpm install` with pnpm version 9.0.6. it reproduce the whole lockfile.
I opened a PR for add `--frozen-lockfile` in #216

lukasjhan:
So I think there are two option. 

1. add `--frozen-lockfile` in ci command.
2. commit new lockfile from new version of pnpm
3. or both

PR #216, opened 2024-04-25
ci: fix pnpm install change lockfile during ci

lukasjhan:
@cre8 About #213, How about trying this change?


lukasjhan:
Or we could commit the lockfile from new version of pnpm

lukasjhan:
> @lukasjhan from my point I would prefer the newer version approach. frozen-lockfile could generate other side effects.

Okay I'll change the pr and let you know :)

PR #225, opened 2024-05-06
Feat/jwt status list

### LINKED ISSUE #224: Feat: Add revocation mechanism to sd-jwt-vc
The spec the referencing a status check when one is provided: https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-01.html#section-3.2.2.2-3.7.1

I implemented a small library that will help to create a statuslist and interact with it and also packs it into a jwt and unpacks it. https://www.npmjs.com/package/jwt-status-list 

What is not covered is the validation of the JWT that is providing the status list. For this point I think the user needs to implement a verification function to validate the `statuslist+jwt` since the public key can be referenced in to many ways.

cre8:
closes #224 

I extended the verify method so all checked that should be done via sd-jwt are called and then we are looking for revocation.

We will only check the revocation if the `status` field is present. We are not throwing an error if somebody has referenced another status mechanism and we are not validating if the status field was correctly placed inside the payload when the credential is created (but types are provided).

The verifier has to implement the fetch and validation function of the jwt-status-token and also the logic how to deal with the status (it can have multiple definitions but the values are not covered by the spec).

The statusValidator function is to stop the current validation. In case the status has a value to continue, the function executes without throwing an error and the verifier has to deal with the state after the successful verification.

### DIFF
--- a/packages/sd-jwt-vc/package.json
+++ b/packages/sd-jwt-vc/package.json
@@ -20,7 +20,11 @@
     "test:e2e": "vitest run ./test/*e2e.spec.ts --environment node",
     "test:cov": "vitest run --coverage"
   },
-  "keywords": ["sd-jwt", "sdjwt", "sd-jwt-vc"],
+  "keywords": [
+    "sd-jwt",
+    "sdjwt",
+    "sd-jwt-vc"
+  ],
   "engines": {
     "node": ">=18"
   },
@@ -35,22 +39,29 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
-    "@sd-jwt/core": "workspace:*"
+    "@sd-jwt/core": "workspace:*",
+    "jwt-status-list": "^0.0.4"
   },
   "devDependencies": {
     "@sd-jwt/crypto-nodejs": "workspace:*",
-    "@sd-jwt/types": "workspace:*"
+    "@sd-jwt/types": "workspace:*",
+    "jose": "^5.2.2"
   },
   "publishConfig": {
     "access": "public"
   },
   "tsup": {
-    "entry": ["./src/index.ts"],
+    "entry": [
+      "./src/index.ts"
+    ],
     "sourceMap": true,
     "splitting": false,
     "clean": true,
     "dts": true,
-    "format": ["cjs", "esm"]
+    "format": [
+      "cjs",
+      "esm"
+    ]
   },
   "gitHead": "ded40e4551bde7ae93083181bf26bd1b38bbfcfb"
 }
--- a/packages/sd-jwt-vc/package.json
+++ b/packages/sd-jwt-vc/package.json
@@ -20,7 +20,11 @@
     "test:e2e": "vitest run ./test/*e2e.spec.ts --environment node",
     "test:cov": "vitest run --coverage"
   },
-  "keywords": ["sd-jwt", "sdjwt", "sd-jwt-vc"],
+  "keywords": [
+    "sd-jwt",
+    "sdjwt",
+    "sd-jwt-vc"
+  ],
   "engines": {
     "node": ">=18"
   },
@@ -36,7 +40,7 @@
   "license": "Apache-2.0",
   "dependencies": {
     "@sd-jwt/core": "workspace:*",
-    "jwt-status-list": "^0.0.4"
+    "jwt-status-list": "^0.0.5"
   },
   "devDependencies": {
     "@sd-jwt/crypto-nodejs": "workspace:*",
@@ -47,12 +51,17 @@
     "access": "public"
   },
   "tsup": {
-    "entry": ["./src/index.ts"],
+    "entry": [
+      "./src/index.ts"
+    ],
     "sourceMap": true,
     "splitting": false,
     "clean": true,
     "dts": true,
-    "format": ["cjs", "esm"]
+    "format": [
+      "cjs",
+      "esm"
+    ]
   },
   "gitHead": "ded40e4551bde7ae93083181bf26bd1b38bbfcfb"
 }
--- a/packages/sd-jwt-vc/package.json
+++ b/packages/sd-jwt-vc/package.json
@@ -20,11 +20,7 @@
     "test:e2e": "vitest run ./test/*e2e.spec.ts --environment node",
     "test:cov": "vitest run --coverage"
   },
-  "keywords": [
-    "sd-jwt",
-    "sdjwt",
-    "sd-jwt-vc"
-  ],
+  "keywords": ["sd-jwt", "sdjwt", "sd-jwt-vc"],
   "engines": {
     "node": ">=18"
   },
@@ -51,17 +47,12 @@
     "access": "public"
   },
   "tsup": {
-    "entry": [
-      "./src/index.ts"
-    ],
+    "entry": ["./src/index.ts"],
     "sourceMap": true,
     "splitting": false,
     "clean": true,
     "dts": true,
-    "format": [
-      "cjs",
-      "esm"
-    ]
+    "format": ["cjs", "esm"]
   },
   "gitHead": "ded40e4551bde7ae93083181bf26bd1b38bbfcfb"
 }
--- a/packages/sd-jwt-vc/src/index.ts
+++ b/packages/sd-jwt-vc/src/index.ts
@@ -47,15 +47,26 @@ export class SDJwtVcInstance extends SDJwtInstance<SdJwtVcPayload> {
     }
   }
 
-  /**
-   * Fetches the status list from the uri.
-   * @param uri
-   * @returns compact JWT
-   */
-  private async statusListFetcher(uri: string): Promise<string> {
-    // according to the spec we assume according to the spec that the response is a compact JWT
-    return fetch(uri).then((res) => res.text());
+/**
+ * Fetches the status list from the uri with a timeout of 10 seconds.
+ * @param uri The URI to fetch from.
+ * @returns A promise that resolves to a compact JWT.
+ */
+private async statusListFetcher(uri: string): Promise<string> {
+  const controller = new AbortController();
+  const timeoutId = setTimeout(() => controller.abort(), 10000);
+
+  try {
+    const response = await fetch(uri, { signal: controller.signal });
+    if (!response.ok) {
+      throw new Error(`Error fetching status list: ${response.status} ${await response.text()}`);
+    }
+
+    return response.text();
+  } finally {
+      clearTimeout(timeoutId);
   }
+}
 
   /**
    * Validates the status, throws an error if the status is not 0.
--- a/packages/sd-jwt-vc/src/index.ts
+++ b/packages/sd-jwt-vc/src/index.ts
@@ -2,14 +2,22 @@ import { SDJwtInstance } from '@sd-jwt/core';
 import type { DisclosureFrame } from '@sd-jwt/types';
 import { SDJWTException } from '../../utils/dist';
 import type { SdJwtVcPayload } from './sd-jwt-vc-payload';
-
-export { SdJwtVcPayload } from './sd-jwt-vc-payload';
-
+import { getListFromStatusListJWT } from 'jwt-status-list';
+import type { SDJWTVCConfig } from './sd-jwt-vc-config';
 export class SDJwtVcInstance extends SDJwtInstance<SdJwtVcPayload> {
   /**
    * The type of the SD-JWT-VC set in the header.typ field.
    */
   protected type = 'vc+sd-jwt';
+
+  protected userConfig: SDJWTVCConfig = {};
+
+  constructor(userConfig?: SDJWTVCConfig) {
+    super(userConfig);
+    if (userConfig) {
+      this.userConfig = userConfig;
+    }
+  }
 
   /**
    * Validates if the disclosureFrame contains any reserved fields. If so it will throw an error.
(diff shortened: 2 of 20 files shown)


packages/sd-jwt-vc/package.json
```diff
@@ -35,11 +35,13 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
-    "@sd-jwt/core": "workspace:*"
+    "@sd-jwt/core": "workspace:*",
+    "jwt-status-list": "^0.0.4"
```
berendsliedrecht, line 39: Love the addition! I think to do a full review of this we should also review the `jwt-status-list` repository. 

Is that something you would also consider donating to this repository? `@sd-jwt/status-list` would work quite well, even if it is not fully sd-jwt specific.

Also, I see that the `jwt-status-list` has a direct dependency on `jose`, which IIRC does not work in React Native. Due to this, Credo would not be able to use this library anymore. I think we should either remove the `jose` dependency or make the status list functionality plugable and the user providers the dependency for their use case.

cre8: I mentioned the "home" of the `jwt-status-list` in the OWF discord channel. I would be fine to move it for now to the monorepo of `@sd-jwt` since it is primarily designed for this.

The requirement of JOSE should be removed when it's blocking the integration of react native. But before removing it and try to implement everything on our own I would love to get the feedback that jose is not compatible with react native (since jose has no dependencies)

berendsliedrecht: > I mentioned the "home" of the `jwt-status-list` in the OWF discord channel. I would be fine to move it for now to the monorepo of `@sd-jwt` since it is primarily designed for this.

Perfect!

> The requirement of JOSE should be removed when it's blocking the integration of react native. But before removing it and try to implement everything on our own I would love to get the feedback that jose is not compatible with react native (since jose has no dependencies)

There are only [two runtimes](https://github.com/panva/jose/tree/2595451b67772dfc82231cddd43050293857139f/src/runtime) (browser and nodejs), which both use node.js or browser specific APIs. This could be fixed with a bunch of polyfills, but I think that the status list library does not need to do any heavy JWT functionality and can be implemented manually. Does not have to be in this PR, but we would need to do it before integrating it within Credo. I also would not mind helping with it.

TimoGlastra:
Should this instead be using https://datatracker.ietf.org/doc/draft-ietf-oauth-status-list/?

That's the replacement of JWT / CWT status list i believe.

packages/sd-jwt-vc/package.json
cre8: @berendsliedrecht I published version 0.0.5 without the dep of `jose`, it was moved to dev deps for the testing.

TimoGlastra: Also agree with Berend, it would be great to not depend on jose, and keep everything platform/runtime agnostic. Polyfills for crypto are the biggest pain of RN.

If the implementation is moved to this repo, we can reuse the crypto callbacks used for signing/verifying SD-JWT VC?

cre8:
> Should this instead be using https://datatracker.ietf.org/doc/draft-ietf-oauth-status-list/?
> 
> That's the replacement of JWT / CWT status list i believe.

It was built on the referenced IETF. I also opened a PR in the sd-jwt-vc spec since it was pointing to the wrong file

packages/sd-jwt-vc/src/index.ts
```diff
@@ -34,4 +42,47 @@ export class SDJwtVcInstance extends SDJwtInstance<SdJwtVcPayload> {
...
+      .verify(encodedSDJwt, requiredClaimKeys, requireKeyBindings)
+      .then((res) => {
+        return { payload: res.payload as SdJwtVcPayload, header: res.header };
+      });
+
+    if (result.payload.status) {
+      //checks if a status field is present in the payload based on https://www.ietf.org/archive/id/draft-ietf-oauth-status-list-02.html
+      if (result.payload.status.status_list) {
+        // fetch the status list from the uri
+        result.payload.status.status_list.uri;
```
TimoGlastra, line 65:
```suggestion
```

packages/sd-jwt-vc/src/sd-jwt-vc-status-reference.ts
```diff
@@ -0,0 +1,9 @@
+export interface SDJWTVCStatusReference {
+  // REQUIRED. implenentation according to https://www.ietf.org/archive/id/draft-ietf-oauth-status-list-02.html
```
TimoGlastra, line 2: Should we use zod or something to validate the structure of this?

packages/sd-jwt-vc/package.json
cre8: The creation of the signed jwt is out of scope of the library. You get the payload and header and then decide yourself how you want to issue the jwt that holds the status list. For the verification the callbacks are provided + added tests

packages/sd-jwt-vc/src/test/index.spec.ts
```diff
@@ -53,3 +76,108 @@ describe('App', () => {
...
+describe('Revocation', () => {
+  const { signer, verifier } = createSignerVerifier();
+  const sdjwt = new SDJwtVcInstance({
+    signer,
+    signAlg: 'EdDSA',
+    verifier,
+    hasher: digest,
+    hashAlg: 'SHA-256',
+    saltGenerator: generateSalt,
+    statusListFetcher(uri: string) {
```
TimoGlastra, line 89: Can't we model this a bit more like we do with SD-JWT, so you provide a signer and verification callback for example. To me it's a bit hidden that the statuListFetcher must aslo verify the status list. If we move this to the sd-jwt repo we could reuse verification callback and there's the following steps:
- call fetcher (provided by user)
- decode jwt, validate structure 
- verify signature (provided by user)


Also for the fetcher in the future it might be nice to provide a default implementation based on fetch. Starting from node 18.something all environments (Node, Browser, React Native) now implement a global fetch we could use?

cre8:
To make a clean PR, how should we proceed? Merge this PR and then open another to transfer the Lib directly into the repo.

Or do all in one step? I would prefer the first option since it makes a cleaner cut between the two tasks

packages/sd-jwt-vc/src/test/index.spec.ts
```diff
@@ -53,3 +76,108 @@ describe('App', () => {
...
+        status_list: {
+          uri: 'https://example.com/status-list',
+          idx: 1,
+        },
+      },
+    };
+    const expectedPayload: SdJwtVcPayload = { iat, iss, vct, ...claims };
+    const encodedSdjwt = await sdjwt.issue(expectedPayload);
+    const result = sdjwt.verify(encodedSdjwt);
+    expect(result).rejects.toThrowError('Status is not valid');
```
TimoGlastra, line 129: As also mentioned in the Credo PR to @lukasjhan, it would be great if the verify method could return an object of validations (as we had in our sd-jwt implementation before merging them together).

For us it's relevant to know which validations succeed and which don't (sometimes it's ok that a cred is revoked). Sometimes it's ok that it's expired. It would be nice to not throw an error but return an validation object

packages/sd-jwt-vc/src/index.ts
```diff
@@ -34,4 +42,47 @@ export class SDJwtVcInstance extends SDJwtInstance<SdJwtVcPayload> {
...
+        );
+        // get the status list from the status list JWT
+        const statusList = getListFromStatusListJWT(statusListJWT);
+        const status = statusList.getStatus(
+          result.payload.status.status_list.idx,
+        );
+        // validate the status
+        if (!this.userConfig.statusValidator) {
+          throw new SDJWTException('Status validator not found');
+        }
```
TimoGlastra, line 81: Can we have a default status list implementation? There's some values defined in the base specification which I think will cover 99% of cases, so most users will never have to provide a custom validator i think

TimoGlastra (commented):
Nice work @cre8, we were intending to start work on this later this month, and left some comments based on how we wanted to approach it

packages/sd-jwt-vc/src/index.ts
lukasjhan: That's a good idea. The default one makes it much easier to use

packages/sd-jwt-vc/src/test/index.spec.ts
lukasjhan: @cre8 @berendsliedrecht 
What do you guys think? 

If we're going to change it, I think it would be better to go through a breaking change before v1.0.

packages/sd-jwt-vc/src/index.ts
lukasjhan: I think It's okay to make one after this PR :)

packages/sd-jwt-vc/src/test/index.spec.ts
berendsliedrecht: I agree with @TimoGlastra. We have had this pattern provided by some previous libraries as well and it works quite nice. As Timo mentioned, some validation failures are okay.

packages/sd-jwt-vc/src/sd-jwt-vc-status-reference.ts
berendsliedrecht: I also would not be opposed to introduce zod into this library and heavily rely on it. It offers some very easy to use validation and parsing which should help a lot in general.

berendsliedrecht:
> @berendsliedrecht @TimoGlastra To make a clean PR, how should we proceed? Merge this PR and then open another to transfer the Lib directly into the repo.
> 
> Or do all in one step? I would prefer the first option since it makes a cleaner cut between the two tasks

First option is fine for me.

cre8:
> > @berendsliedrecht @TimoGlastra To make a clean PR, how should we proceed? Merge this PR and then open another to transfer the Lib directly into the repo.
> > Or do all in one step? I would prefer the first option since it makes a cleaner cut between the two tasks
> 
> First option is fine for me.

Well, too late :D

packages/sd-jwt-vc/src/index.ts
TimoGlastra: I don't see why we need to make the simple things harder because we're not covering all cases? If you do the complex flow you probably want to do something custom anyway, but if you're just getting started having defaults is really helpful.

I don't see a reason why not. 

(Maybe we need to have some document where we define the goals and non-goals of this library. A lot of the discussions we're having in this PR seems to be revolved around assumptions in what it should do and how things should work)

My assumption:
- cross platform (including react native)
- easy to get started, customizable if you need it / sane defaults (I think we may be misaligned on this topic @cre8?)
- ..?

packages/sd-jwt-vc/src/sd-jwt-vc-status-reference.ts
cre8: I opened #227 for this

packages/sd-jwt-vc/src/test/index.spec.ts
cre8: A default fetcher is provided as a fallback, the fetch function. It is supported by node v18 and web browsers and since this lib is >=18 this should be fine

packages/sd-jwt-vc/src/test/index.spec.ts
cre8: Opened #228 for this

packages/sd-jwt-vc/src/index.ts
cre8: I added default implementations for fetcher and status validation that can be overwritten.

lukasjhan (approved):

packages/sd-jwt-vc/src/index.ts
lukasjhan: I thought you removed this line :)

packages/sd-jwt-vc/src/index.ts
```diff
@@ -34,4 +46,79 @@ export class SDJwtVcInstance extends SDJwtInstance<SdJwtVcPayload> {
...
+
+  /**
+   * Fetches the status list from the uri.
+   * @param uri
+   * @returns compact JWT
+   */
+  private async statusListFetcher(uri: string): Promise<string> {
+    // according to the spec we assume according to the spec that the response is a compact JWT
+    return fetch(uri).then((res) => res.text());
+  }
```
TimoGlastra, line 58: I think it would be good to add a timeout to this. It would also be good to e.g. add handling for large response size (so if the body exceeds 1MB we will cancel the fetch to prevent a very large payload from crashing the system). But that can be added later / is maybe better suited for a custom implementation (but it does affect security so that's something that should be taken into account.

Also the method didn't take into account non 200 responses in which case response.text is probably not a JWT

```suggestion
/**
 * Fetches the status list from the uri with a timeout of 10 seconds.
 * @param uri The URI to fetch from.
 * @returns A promise that resolves to a compact JWT.
 */
private async statusListFetcher(uri: string): Promise<string> {
  const controller = new AbortController();
  const timeoutId = setTimeout(() => controller.abort(), 10000);

  try {
    const response = await fetch(uri, { signal: controller.signal });
    clearTimeout(timeoutId);
    if (!response.ok) {
      throw new Error(`Error fetching status list: ${response.status} ${await response.text()}`);
    }
    
    return await response.text();
  } catch (error) {
    clearTimeout(timeoutId);
    throw error;
  }
}
```

TimoGlastra: Yeah agreed! LGTM

TimoGlastra (approved):
Really nice work @cre8. LGTM ✅ 

I'm going to implement this into Credo directly, and will let you know if I encounter any issues

cre8:
@TimoGlastra new version `0.7.0` got released.

PR #237, opened 2024-05-20
Fix/status list

cre8:
Add a getter for the bitstring


packages/jwt-status-list/src/status-list.ts
```diff
@@ -31,6 +31,14 @@ export class StatusList {
     this.totalStatuses = statusList.length;
   }
 
+  /**
+   * Get the status list.
+   * @returns
```
berendsliedrecht, line 36: This misses the type. Or it should be removed.

packages/jwt-status-list/src/status-list.ts
```diff
@@ -31,6 +31,14 @@ export class StatusList {
...
+   */
+  getStatusList(): number[] {
```
berendsliedrecht, line 38: Why not an actual `get`?

cre8: Updates to get, good point!

packages/jwt-status-list/src/status-list.ts
cre8: returns got removed, also for the other functions. Since each function has a typed return, it should be enough.

lukasjhan (approved):
Looks good to me :)

PR #240, opened 2024-05-21
Cre8/issue239

### LINKED ISSUE #239: Format files after release
When lerna is publishing a new release, it will update the `package.json` files and push it to github. Unfortunately the changes are not aligned with our biome config so the CI run after this will fail, see https://github.com/openwallet-foundation-labs/sd-jwt-js/actions/runs/9170609357/job/25213207873

There are two options:
- Update biome config: find out how to configure the linting rules so it is compliant to the formatting lerna is doing
- run biome before pushing to github. For this we will pass the  `--no-push` to lerne, run the biome lint command, commit the changes and push them to main.

cre8:
fixes #239 

I am not sure if we can test it locally since we do not have a valid npm token to run the task successfully. Based on the logic:
- lerna publishes to npm, creates the commits but does not push
- we run the format command and commit the changes
- Then we run execute the push to push the two commits.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -183,3 +183,10 @@ jobs:
 
       - name: 'Publish latest version'        
         run: pnpm publish:latest
+
+      - name: 'Format generated files and commit changes'
+        run: | 
+          pnpm format
+          git add .
+          git commit -m "chore: format generated files"
+          git push origin HEAD:main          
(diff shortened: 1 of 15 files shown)


lukasjhan:
Yeah.. we can't test it locally. Let's test it when we have next release :)

berendsliedrecht (approved):

lukasjhan (approved):
LGTM :)

PR #243, opened 2024-07-18
Fix KeyBinding not being propogated bug

orenyomtov:
The `kb` property is not being properly propagated in the `verify` function

### DIFF
--- a/packages/sd-jwt-vc/src/sd-jwt-vc-instance.ts
+++ b/packages/sd-jwt-vc/src/sd-jwt-vc-instance.ts
@@ -104,7 +104,7 @@ export class SDJwtVcInstance extends SDJwtInstance<SdJwtVcPayload> {
     const result = await super
       .verify(encodedSDJwt, requiredClaimKeys, requireKeyBindings)
       .then((res) => {
-        return { payload: res.payload as SdJwtVcPayload, header: res.header };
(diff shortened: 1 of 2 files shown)


cre8 (approved):

cre8:
@orenyomtov thank you for bringing this up. It seems we have overseen it since the property is optional.

lukasjhan (approved):
Thank you :)

orenyomtov:
Wow what a fast review and merge!

When can the next package release including this fix can be expected?

orenyomtov:
Hey @lukasjhan, can you release a new version of the package so I can use the patch?

lukasjhan:
@orenyomtov Sorry didn't notify you about the release.
It was released at v0.7.2.

lukasjhan:
Oh The tag wasn't created. We actually release right after the merge.
Sorry you must have waited 13 days :)

orenyomtov:
oh cool, tnx!

cre8:
@lukasjhan I think because we updated the release approach. Lerna is not pushing the created files because we wanted to format them before pushing. So I think the files get pushed, but not the created tags

lukasjhan:
@cre8 Oh I see. Thank you for the fix. I was looking for the lerna's configuration 😢

PR #244, opened 2024-07-30
fix: check if the header includes the string

cre8:
Before we checked for equal, ignoring that the charset will be added. Using include solves this problem


lukasjhan (approved):
Looks good to me :)

PR #245, opened 2024-08-01
fix: also push tags when releasing

cre8:
This fix should also push the tags that are required for a new release to github


lukasjhan (approved):
Looks good to me :)

PR #247, opened 2024-08-16
Feat/type metadata

cre8:
Adding the schema validation based on the [typed metadata](https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-04.html#name-type-metadata).

This feature has to be activated with a parameter when the instance is created without breaking existing applications. The function to fetch the typed metadata can be overridden, the default one with fetch it from the url.

At this point it is only used for validation, after this [MR](https://github.com/oauth-wg/oauth-sd-jwt-vc/pull/236) got merged, the metadata should be fetched via a function (either be returned after the validation and/or it's able to fetch them via a dedicated call to skip the validation of the credential) to receive the display information.

### DIFF
--- a/packages/sd-jwt-vc/src/sd-jwt-vc-instance.ts
+++ b/packages/sd-jwt-vc/src/sd-jwt-vc-instance.ts
@@ -1,13 +1,23 @@
 import { Jwt, SDJwtInstance } from '@sd-jwt/core';
-import type { DisclosureFrame, Verifier } from '@sd-jwt/types';
+import type { DisclosureFrame, Hasher, Verifier } from '@sd-jwt/types';
 import { SDJWTException } from '@sd-jwt/utils';
 import type { SdJwtVcPayload } from './sd-jwt-vc-payload';
-import type { SDJWTVCConfig } from './sd-jwt-vc-config';
+import type {
+  SDJWTVCConfig,
+  StatusListFetcher,
+  StatusValidator,
+} from './sd-jwt-vc-config';
 import {
   type StatusListJWTPayload,
   getListFromStatusListJWT,
+  type StatusListJWTHeaderParameters,
 } from '@sd-jwt/jwt-status-list';
-import type { StatusListJWTHeaderParameters } from '@sd-jwt/jwt-status-list';
+import type { TypeMetadataFormat } from './sd-jwt-vc-type-metadata-format';
+import Ajv, { type SchemaObject } from 'ajv';
+import type { VerificationResult } from './verification-result';
+import addFormats from 'ajv-formats';
+import type { VcTFetcher } from './sd-jwt-vc-vct';
+
 export class SDJwtVcInstance extends SDJwtInstance<SdJwtVcPayload> {
   /**
    * The type of the SD-JWT-VC set in the header.typ field.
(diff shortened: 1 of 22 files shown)


packages/sd-jwt-vc/src/sd-jwt-vc-instance.ts
```diff
@@ -113,12 +123,167 @@ export class SDJwtVcInstance extends SDJwtInstance<SdJwtVcPayload> {
...
+  /**
+   * Fetches the content from the url with a timeout of 10 seconds.
+   * @param url
+   * @returns
+   */
+  private async fetch<T>(url: string, integrity?: string) {
+    const controller = new AbortController();
+    const timeoutId = setTimeout(() => controller.abort(), 10000);
+    try {
+      const response = await fetch(url, {
```
lukasjhan, line 190: Oh RN also support the fetch function. Good :)

packages/sd-jwt-vc/src/sd-jwt-vc-instance.ts
```diff
@@ -113,12 +123,167 @@ export class SDJwtVcInstance extends SDJwtInstance<SdJwtVcPayload> {
...
+    //init the json schema validator, load referenced schemas on demand
+    const schema = await this.loadSchema(typeMetadataFormat);
+    const loadedSchemas = new Set<string>();
+    // init the json schema validator
+    const ajv = new Ajv({
+      loadSchema: async (uri: string) => {
+        if (loadedSchemas.has(uri)) {
+          return {};
+        }
+        const response = await fetch(uri);
```
lukasjhan, line 250:
```suggestion
        const response = await this.fetch(uri);
```
Should it be `this.fetch`?

packages/sd-jwt-vc/src/sd-jwt-vc-instance.ts
cre8: Since fetch is supported in the browser and in node, we can use it for both. I see no point that we need it to override with a custom implementation (at this point)

packages/sd-jwt-vc/src/sd-jwt-vc-instance.ts
cre8: It does not really matter. The internal fetch function has an integrity validation. This is not passed when loading schemas. However I will check if schemas have an integrity validation and if we can add it then like we are doing it with the other resources from vct

packages/sd-jwt-vc/src/sd-jwt-vc-instance.ts
lukasjhan: Yeah I agree.

lukasjhan (approved):

packages/sd-jwt-vc/src/sd-jwt-vc-instance.ts
cre8: @lukasjhan I found no standard for validating the integrity of a referenced $schema value. At this point I do not see any privacy concerns when we are not validating it for now. In case we see something, we can open it.

packages/sd-jwt-vc/src/sd-jwt-vc-instance.ts
cre8: fetch is also used in the statuslist fetcher, but the whole method can be overwritten via userconfig. So I think it's fine for now. Otherwise we need to implement a general fetch function that is covering all cases.

PR #249, opened 2024-08-20
Add serialization for json and json flattened

### LINKED ISSUE #242: Support JWS JSON Serialization
In [sd-jwt draft 09 section 9](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-09.html#name-jws-json-serialization.), JSON Serialization method is added. Let's support this feature in our library

cre8:
closes #242 

Since the general serialization function allows to add multiple signature, we need a new internal structure to store the information in an sd-jwt instance (right now it is managed as compact json).

This PR should introduce the serialization of compact, json or flattened json. compact is the default value to not break existing implementations.

It should be the goal to issue and present an sd-jwt as one of the three formats and all three formats should be used as an input.

I also added some tests to validate the encoding, decoding, but the expected value was generated based on the own function so it would help if someone could challenge the algorithms with the current spec.

@lukasjhan I would suggest to store the elements as an object inside the sd-jwt like the json presentation (so signatures are managed in an array)

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -2,8 +2,8 @@ import { SDJWTException, uint8ArrayToBase64Url } from '@sd-jwt/utils';
 import { Jwt } from './jwt';
 import { KBJwt } from './kbjwt';
 import {
-  SDJJWTJson,
-  SDJWTJsonFlattened,
+  type SDJJWTJson,
+  type SDJWTJsonFlattened,
   SDJwt,
   pack,
   type SDJWTType,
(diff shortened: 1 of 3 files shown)


cre8:
@lukasjhan in case of signature validation, we need to update the internal storage of signature in the `JWT` object: https://github.com/openwallet-foundation-labs/sd-jwt-js/blob/main/packages/core/src/jwt.ts#L117 Because for the json approach there could be multiple signatures.
I think it's better to update the signature from a string to an array instead of building multiple JWT instances and manage them somewhere else

cre8:
Closing this in favor of #253

PR #252, opened 2024-11-01
Update example/ all.ts

robinMcA:
Uses the wrong jwt, presents the private Cred when it should use the public presentation


cre8 (approved):
LGTM, thank you @robinMcA for the fix!

PR #253, opened 2024-11-11
feat: add jws serialization feature

lukasjhan:
This PR implements [new JWS JSON serialization](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-13.html#name-jws-json-serialization) in standard.

Currently, I implemented the converting features. If you want, I can implement issue, present, and verify feature additionally on it. It may have dup code but I think it's quite simple approach for the first version since It doesn't ruin the original code. 

Perhaps I did a naive implementation or I might be wrong because I did not know the intended use cases of the standard. 
If there is anything you would like to point out, please feel free to do so. Thank you :)

### DIFF
--- a/examples/sd-jwt-example/flattenJSON.ts
+++ b/examples/sd-jwt-example/flattenJSON.ts
@@ -1,0 +1,57 @@
+import { FlattenJSON, SDJwtInstance } from '@sd-jwt/core';
+import type { DisclosureFrame } from '@sd-jwt/types';
+import { createSignerVerifier, digest, generateSalt, ES256 } from './utils';
+
+(async () => {
+  const { signer, verifier } = await createSignerVerifier();
+
+  // Create SDJwt instance for use
+  const sdjwt = new SDJwtInstance({
+    signer,
+    signAlg: ES256.alg,
+    verifier,
+    hasher: digest,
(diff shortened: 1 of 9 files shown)


cre8:
I like that the flattened approach is realized with some extra functions. However for the json format where there are multiple signatures, this approach will not work anymore. 
So is it good now to add it and later update it or better now?
For JADES signatures we will need the generalized approach as object

lukasjhan:
> I like that the flattened approach is realized with some extra functions. However for the json format where there are multiple signatures, this approach will not work anymore. So is it good now to add it and later update it or better now? For JADES signatures we will need the generalized approach as object

I'll implement issue, present, verify feature for GeneralJSON to finish my pr :)

lukasjhan:
@cre8 I added separated instance class for issue, present and verify for generalJSON.

cre8 (approved):

lukasjhan:
> @lukasjhan changes are looking good for me. I would prefer a more object oriented approach (and I think we have some redundancy), but it should be fine for now without a risk of breaking changes

Thank you :)  I'm merging this to start working on the examples and tests.
Let's reduce redundant code after standard finalized.

PR #256, opened 2024-11-15
feat: align media type with sd-jwt-vc draft 06

charsleysa:
Draft 06 changes the media type from `vc+sd-jwt` to `dc+sd-jwt`.
https://author-tools.ietf.org/iddiff?url1=draft-ietf-oauth-sd-jwt-vc-05&url2=draft-ietf-oauth-sd-jwt-vc-06&difftype=--html
https://github.com/oauth-wg/oauth-sd-jwt-vc/pull/268


cre8:
@charsleysa thanks for the PR!

@lukasjhan I would be fine with this change without marking it as a breaking change. The sd-jwt vc spec is still in draft

lukasjhan (approved):
LGTM. Thank you

lukasjhan:
@cre8 I'm also fine with this change :). Let's merge it

PR #260, opened 2024-12-10
feat: align type metadata fetching with sd-jwt-vc draft 08

charsleysa:
Draft 07 removes the `.well-known/vct/` addition to the `vct` URI when fetching the Type Metadata. Draft 08 is the latest.


cre8:
Thank you for the update @charsleysa 
Have you checked if there are any other updates for the latest change that we need to be aware of?

cre8 (approved):

lukasjhan (approved):
LGTM Thank you

PR #262, opened 2024-12-13
fix: push changes

cre8:
Let lerna push directly, biome config needs to be updated later


lukasjhan (approved):
LGTM

PR #266, opened 2025-01-12
chore: lowercase sha algorithm to be compliant

### LINKED ISSUE #264: Hash algorithm names should be written in lower case
This repository seems to use mixed case characters when defining hash algorithm names, see search results https://github.com/search?q=repo%3Aopenwallet-foundation%2Fsd-jwt-js+%22SHA-%22&type=code. Specifically the online tool https://www.sdjwt.co/issue seems to be fixed to use sha-256 algorithm to calculate the SD-JWT Disclosure hashes, but it outputs the algorithm name to the SD-JWT payload in upper case as `"_sd_alg": "SHA-256"`. This is incorrect. The hash algorithm names are defined in [IANA Named Information Hash Algorithm Registry](https://www.iana.org/assignments/named-information/named-information.xhtml) and they should be treated as case-sensitive. See also https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-14.html#section-4.1.1-2.

cre8:
Closes #264

Renaming `SHA-256` to `sha-256` to be compliant with iana naming as described in the issue.

I also added the type based on the iana values so autocomplete should warn the user when they use a wrong value and we are checking it during issuance (we do not this during decoding in case the sd jwt was created by another tool, ignoring these rules)

### DIFF
--- a/docs/0.x/sdjwt-instance.md
+++ b/docs/0.x/sdjwt-instance.md
@@ -10,7 +10,7 @@ const sdjwt = new SDJwtInstance({
   verifier,
   signAlg: 'EdDSA',
   hasher: digest,
-  hashAlg: 'SHA-256',
+  hashAlg: 'sha-256',
   saltGenerator: generateSalt,
 });
 ```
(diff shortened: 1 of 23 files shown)


lukasjhan (approved):
LGTM, There might be an update burden related to the type. but I think the impact will be very small.

Issue #258, opened 2024-11-19
Not being able to fetch type metadata should not fail verification?


TimoGlastra:
As I understand the SD-JWT VC specification, the `vct` CAN be used to represent the type metadata URL, but i don't think it's invalid to use `https://` uri for vct and not host type metadata?

Also the error message is the response of the VCT, which can be confusing. Instead of using `response.text()` as the error message, maybe it should be `Error fetching type metadata. ${await response.text()}`?

cre8:
Hi @TimoGlastra ,

you can pass a parameter if the metadata should be loaded from the vct by setting a config value: https://github.com/openwallet-foundation/sd-jwt-js/blob/main/packages/sd-jwt-vc/src/sd-jwt-vc-config.ts#L18
the default value is false, so when there is a url it will not be fetched. But when there is another url and you set it to fetching, it will end in an error.

For the error for schema fetching we have this error message: https://github.com/openwallet-foundation/sd-jwt-js/blob/main/packages/sd-jwt-vc/src/sd-jwt-vc-instance.ts#L252

But in the fetch function it's generic in the `fetch` function because this function is used to fetch the status list and the schema that are throwing case specific error messages.

TimoGlastra:
The config option assumes you know upfront whether the SD-JWT has type metadata. I'd like to fetch the type metadata if it exists, but if the endpoint returns 404 that seems fine to me.

cre8:
> The config option assumes you know upfront whether the SD-JWT has type metadata. I'd like to fetch the type metadata if it exists, but if the endpoint returns 404 that seems fine to me.

When you are activating the `loadMetadataFormat`, from where should you get the schema to validate it when not from a successful response? If you set it to true and the function is not able to load it, there is no other way to run the validation.

You are able to overwrite the `vctFether` via the config and so implement your own logic. For me the normal way sounds logic that when somebody adds a URL, starting with `http://` it's meant to be fetchable. This allows you for example to look up a a database for the schema, rather then fetching it from a URL.

So when you want to perform schema validation and you know you will get other values than URL in the vct that, you need to overwrite this function by passing it.

PR #267, opened 2025-01-13
chore: improve error handing for fetch function

cre8:
Improve the fetch function based on the input from #258 + adding with timeout case and allowing to adjust the timeout value

### DIFF
--- a/packages/sd-jwt-vc/src/sd-jwt-vc-instance.ts
+++ b/packages/sd-jwt-vc/src/sd-jwt-vc-instance.ts
@@ -166,18 +166,32 @@ export class SDJwtVcInstance extends SDJwtInstance<SdJwtVcPayload> {
    * @param url
    * @returns
    */
-  private async fetch<T>(url: string, integrity?: string) {
+  private async fetch<T>(url: string, integrity?: string): Promise<T> {
     const controller = new AbortController();
-    const timeoutId = setTimeout(() => controller.abort(), 10000);
+    const timeoutId = setTimeout(
+      () => controller.abort(),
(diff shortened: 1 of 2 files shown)


packages/sd-jwt-vc/src/sd-jwt-vc-instance.ts
```diff
@@ -166,18 +166,31 @@ export class SDJwtVcInstance extends SDJwtInstance<SdJwtVcPayload> {
...
-        throw new Error(await response.text());
+        const errorText = await response.text();
+        throw new Error(
+          `Error fetching ${url}: ${response.status} ${response.statusText} - ${errorText}`,
+        );
       }
       await this.validateIntegrity(response.clone(), url, integrity);
       return response.json() as Promise<T>;
+    } catch (error) {
+      if (error instanceof DOMException && error.name === 'AbortError') {
```
berendsliedrecht, line 190: In node.js or react native environments, is this still a `DOMException`?

cre8: @berendsliedrecht @lukasjhan I think we should write a test and find it out. I added this instance of check, otherwise the linter would fail because of the missing type.

I will open a new PR to add tests and to correct it

cre8: @berendsliedrecht I ran the test in node `pnpm run test:node` and it threw me the DomException. I am not sure how it will work with react native since we have no chance here to test it...

One solution would be to replace fetch via a http library like axios. But this will add another external lib

PR #268, opened 2025-01-13
type: define IANA hash algorithm list once

srosenda:
Noticed that PR #266 can be improved a bit. It is possible to utilise Typescript's [const assertion](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions) and derive the `HashAlgorithm` type from the `IANA_HASH_ALGORITHMS` array, making it unnecessary to duplicate the algorithm list for both the type and the run-time array. This requires moving the `IANA_HASH_ALGORITHMS` array to the `types` package and exporting it from there.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -17,6 +17,7 @@ import {
   type SDJWTConfig,
   type JwtPayload,
   type Signer,
+  IANA_HASH_ALGORITHMS,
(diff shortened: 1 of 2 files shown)


cre8 (approved):
@srosenda thank you, I implemented first the type and then the array and just merged it :D

PR #269, opened 2025-01-15
Fetch-timeout

cre8:
Set the newer version of the abort signal, also add a test for it

Addressing the discussion from #267 

### DIFF
--- a/examples/sd-jwt-example/generalJSON.ts
+++ b/examples/sd-jwt-example/generalJSON.ts
@@ -37,7 +37,7 @@ import { createSignerVerifier, digest, generateSalt, ES256 } from './utils';
   console.log('encodedSdjwt:', encodedSdjwt);
 
   const generalJSON = GeneralJSON.fromEncode(encodedSdjwt);
-  console.log('flattenJSON(credential): ', generalJSON.toJson());
+  console.log('generalJSON(credential): ', generalJSON.toJson());
 
   const presentedSdJwt = await sdjwt.present<typeof claims>(
     encodedSdjwt,


packages/sd-jwt-vc/src/sd-jwt-vc-instance.ts
```diff
@@ -187,12 +180,10 @@ export class SDJwtVcInstance extends SDJwtInstance<SdJwtVcPayload> {
       await this.validateIntegrity(response.clone(), url, integrity);
       return response.json() as Promise<T>;
     } catch (error) {
-      if (error instanceof DOMException && error.name === 'AbortError') {
+      if (error instanceof DOMException && error.name === 'TimeoutError') {
```
lukasjhan, line 183: Is the DOMException okay for other platform? RN or nodejs?

cre8: I ran the test with as browser and as node (provided by the package file). In both cases I got a TimeoutError.

As mentioned I don't have a RN setup so I am unable to test it. A quick check of the [react native github repo](https://github.com/search?q=repo%3Afacebook%2Freact-native%20domexception&type=code) assumes that it should work.

lukasjhan: I guess the DOMException only works in the browsers. Since we got a TimeoutError in the browser, I think we can remove DOMException checking logic here. Then we have no problem about platforms. :)

cre8: But when you remove the `instanceof DOMExeption`, you will get a compile error because error has no type when asking for `error.name`.

lukasjhan: Oh I see, That's the point. 
Umm... In that case, how about using `instanceof Error`?
Actually The `DOMExeption` extends Error.

packages/types/src/type.ts
```diff
@@ -32,7 +32,6 @@ export const IANA_HASH_ALGORITHMS = [
 
 export type HashAlgorithm = (typeof IANA_HASH_ALGORITHMS)[number];
 
-
```
lukasjhan, line 35: I fixed this in separate pr #271

packages/sd-jwt-vc/src/sd-jwt-vc-instance.ts
cre8: I removed the error handling and just updated the test since both environment are throwing different error message so we need to catch both cases

lukasjhan (approved):
LGTM thank you 👍🏻

PR #272, opened 2025-01-29
fix: Frame fails when a type with an optional or nullable field is used

charsleysa:
Types such as the following currently fail when trying to make the nested fields selectively discloseable:

```ts
type PID = {
    given_name: string;
    address?: {
        country: string;
    }
};

const disclosureFrame: DisclosureFrame<PID> = {
    _sd: ['given_name'],
    address: {
        _sd: ['country']
    }
};

```

Signed-off-by: Stefan Charsley <charsleysa@gmail.com>

### DIFF
--- a/packages/types/src/type.ts
+++ b/packages/types/src/type.ts
@@ -172,7 +172,7 @@ type Frame<Payload> = Payload extends Array<infer U>
   : Payload extends Record<string, unknown>
     ? NonNever<
         {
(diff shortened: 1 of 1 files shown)


lukasjhan (approved):
LGTM. Thank you for the PR :)

cre8:
Thank you @charsleysa for the fix! The latest version was released with it some minutes ago.

PR #280, opened 2025-05-28
feat: add sd-jwt with jades support package

lukasjhan:
SD-JWT with JAdES package is moved from https://github.com/openwallet-foundation-labs/sd-jwt-vc-dm

### DIFF
--- a/packages/jades/src/sign.ts
+++ b/packages/jades/src/sign.ts
@@ -67,7 +67,7 @@ export class Sign<T extends Record<string, unknown>> {
     return base64urlEncode(JSON.stringify({ ...this.protectedHeader, kid }));
   }
 
-  private getSignPayload(kid: string): string {
+  public getSignPayload(kid: string): string {
     const encodedProtectedHeader = this.encodedProtectedHeader(kid);
 
     const encodedPayload =
--- a/packages/jades/src/test/sign.spec.ts
+++ b/packages/jades/src/test/sign.spec.ts
@@ -142,7 +142,7 @@ describe('Sign', () => {
     if (alg.startsWith('RS') || alg.startsWith('PS')) {
       signature = signer.sign({
         key: privateKey,
-        padding: (algorithm as any).padding,
+        padding: (algorithm as { padding: number }).padding,
       });
     } else if (alg.startsWith('ES')) {
       signature = signer.sign({
@@ -174,7 +174,7 @@ describe('Sign', () => {
 
       // Create signature (simulating HSM response)
       // In real usage, this would be done by HSM
-      const signingInput = sign['getSignPayload']('test-kid');
+      const signingInput = sign.getSignPayload('test-kid');
       const signature = createTestSignature(signingInput, 'RS256', privateKey);
 
       // Add the signature
@@ -209,7 +209,7 @@ describe('Sign', () => {
       const hash = await sign.getHash('RS256', 'test-kid');
       expect(hash).toBeInstanceOf(Uint8Array);
 
-      const signingInput = sign['getSignPayload']('test-kid');
+      const signingInput = sign.getSignPayload('test-kid');
       const signature = createTestSignature(signingInput, 'RS256', privateKey);
 
       const result = await sign.addSignature(signature, 'test-kid');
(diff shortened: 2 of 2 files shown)


packages/jades/README.md
```diff
@@ -0,0 +1,218 @@
...
+  const certs = parseCerts(certPem);
+  const kid = createKidFromCert(certs[0]);
+
+  const keyPem = fs.readFileSync('./fixtures/private.pem', 'utf-8');
+  const privateKey = createPrivateKey(keyPem);
+
+  await jades
+    .setProtectedHeader({
+      alg: 'RS256',
+      typ: 'jades',
```
cre8, line 56: must the type be jades? I would assume it has to be something like `dc+sd-jwt`

packages/jades/README.md
```diff
@@ -0,0 +1,218 @@
...
+
+## Features
+
+### SD-JWT VCDM Data Model Profile
+
+This library provides interoperability between SD-JWT (Selective Disclosure JWT) and W3C Verifiable Credentials Data Model:
+
+- Issue Verifiable Digital Credentials in SD-JWT VC format while maintaining W3C VCDM compliance
+- Support for Selective Disclosure capabilities
+- Seamless integration with standard VC verification processes
```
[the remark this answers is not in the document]
lukasjhan: Sry I didn't edit the readme. It's about jades only. You are right

packages/jades/src/present.ts
```diff
@@ -0,0 +1,43 @@
...
+    // Convert string to GeneralJSON if needed
+    const generalJsonCredential = getGeneralJSONFromJWSToken(credential);
+
+    // If there are no disclosures, return the credential as is
+    // This prevents errors from the core library when handling credentials without SD claims
+    if (
+      !generalJsonCredential.disclosures ||
+      generalJsonCredential.disclosures.length === 0
+    ) {
+      console.log(
```
cre8, line 29: remove console output (and the whole if statement), or change it to console.debug.

packages/jades/src/verify.ts
```diff
@@ -0,0 +1,99 @@
...
+      const isValid = JWTVerifier.verifySig(
+        data,
+        signatureB64,
+        header.x5c,
+        header.alg,
+      );
+
+      return isValid;
+    } catch (error) {
+      console.error('JWT token verification error:', error);
```
cre8, line 43: I would avoid to print errors to the console, it's just good for debugging, but not for production.

packages/jades/README.md
lukasjhan: No, It's just an example :)

lukasjhan: Okay. good idea

packages/jades/README.md
```diff
@@ -0,0 +1,210 @@
...
+
+  const keyPem = fs.readFileSync('./fixtures/private.pem', 'utf-8');
+  const privateKey = createPrivateKey(keyPem);
+
+  await jades
+    .setProtectedHeader({
+      alg: 'RS256',
+      typ: 'dc+sd-jwt',
+    })
+    .setX5c(certs)
```
cre8, line 84: why do we have an extra function to set the x5c, is this required by the specification? To make it easier, I would just have a function to set the protected and the unprotected header?

lukasjhan: I made the interface like jsonwebtoken :) It looks developer friendly to me

packages/jades/README.md
```diff
@@ -0,0 +1,210 @@
...
+            tstTokens: [
+              {
+                val: 'Base64-encoded RFC 3161 Timestamp Token',
+              },
+            ],
+          },
+        },
+      ],
+    })
+    .sign(privateKey, kid);
```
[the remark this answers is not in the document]
lukasjhan: Oh I see, I didn't think about use cases with HSM. 
Perhaps would signer function as a parameter be better?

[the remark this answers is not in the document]
lukasjhan: I see, I'll add that function :)

cre8:
@lukasjhan can you pls fix the conflict of the lock file pls?

lukasjhan:
@cre8 I resolved the conflict. and I'm working on implementing getting hash function.
Should I implement to support custom signer and verifier?

cre8:
I think a jades package is good in general, but it should be added here: https://github.com/openwallet-foundation-labs/identity-common-ts

@lukasjhan therefore closing it here, but if you think the codebase is good, we can transition it into the other repo

PR #283, opened 2025-06-10
fix: simplify schema validation error message by removing error details

lukasjhan:
validate.errors contains the payload. so I removed the details on the error


cre8:
I would like to reject this change:

As a developer, it helps me to identify why it failed. 
If I do not want to give this information to the user, I can catch the call of the validation function and then throw a generic one.

PR #285, opened 2025-06-16
SD JWT Instances VCT Metadata get separated from verify

### LINKED ISSUE #284: Load vct values and make them available
Right now we are only using the vctFetcher to check if the schema is [aligned with the passed credential](https://github.com/openwallet-foundation/sd-jwt-js/blob/main/packages/sd-jwt-vc/src/sd-jwt-vc-instance.ts#L213).

It would be great to load it so the Wallet would be able to fetch e.g. the display values via the instance.

# Suggestion
We should make the vctFetcher function public, so the wallet MUST not validate the whole credential when it want to get the information.
The question is should we load the vct values on the instance load or on demand. And should there be a parameter that allows to overrule a mismatch e.g. when the integrity check fails.

patatoid:
Closes #284

### DIFF
--- a/packages/sd-jwt-vc/src/sd-jwt-vc-instance.ts
+++ b/packages/sd-jwt-vc/src/sd-jwt-vc-instance.ts
@@ -1,4 +1,4 @@
-import { Jwt, SDJwtInstance } from '@sd-jwt/core';
+import { Jwt, SDJwt, SDJwtInstance } from '@sd-jwt/core';
 import type { DisclosureFrame, Hasher, Verifier } from '@sd-jwt/types';
 import { SDJWTException } from '@sd-jwt/utils';
 import type { SdJwtVcPayload } from './sd-jwt-vc-payload';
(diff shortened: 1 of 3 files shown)


cre8 (approved):

lukasjhan (approved):

packages/sd-jwt-vc/src/sd-jwt-vc-instance.ts
```diff
@@ -260,6 +270,22 @@ export class SDJwtVcInstance extends SDJwtInstance<SdJwtVcPayload> {
     return typeMetadataFormat;
   }
 
+  /**
+   * Fetches VCT Metadata of the SD-JWT-VC. Returns the type metadata format. If the SD-JWT-VC does not contain a vct claim, an error is thrown.
+   * @param result
+   * @returns
+   */
+  private async fetchVct(result): Promise<TypeMetadataFormat> {
```
cre8, line 278:
```suggestion
  private async fetchVct(result): Promise<TypeMetadataFormat | undefined> {
```

cre8: because when throwing an error, you will get no return

cre8:
@patatoid can you please sign your last commit so the DCO check is passing? Then I am able to merge it.

patatoid:
I noticed there is no git global configuration to sign off commits, made an alias to do so.

Hoping it not took to much time for you to merge the changes :pray:

PR #286, opened 2025-06-16
WIP: feat: add missing types

cre8:
Adding [Display Metadata](https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-09.html#name-display-metadata) and [Claim Metadata](https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-09.html#name-claim-metadata) to the library.

@lukasjhan we should discuss how much functions we want to provide to deal with the information, like adding a function to return the rendered SVG, ect.  Right now I only added the types to it.

### DIFF
--- a/packages/sd-jwt-vc/src/sd-jwt-vc-type-metadata-format.ts
+++ b/packages/sd-jwt-vc/src/sd-jwt-vc-type-metadata-format.ts
@@ -1,13 +1,156 @@
 /**
- * https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-08.html#name-type-metadata-format
+ * Logo metadata used in rendering a credential.
+ */
+export type Logo = {
+  /** REQUIRED. A URI pointing to the logo image. */
+  uri: string;
+  /** OPTIONAL. An "integrity metadata" string as described in Section 7. */
+  'uri#integrity'?: string;
+  /** OPTIONAL. A string containing alternative text for the logo image. */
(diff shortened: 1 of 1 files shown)


lukasjhan:
Thank you @cre8. I think adding type is great.
in my point of view, rendering SVG is kind of too much to include in sd-jwt-vc. Perhaps we could make a separate library like `@sd-jwt/svg-rederer-web` and `@sd-jwt/svg-renderer-rn` if you want :)

lukasjhan:
> so one relevant function would be when claim values are provided, that the credential is following the structure (should maybe also covered by using the json schema when provided) and checking the selective disclosure as described [here](https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-09.html#name-claim-selective-disclosure-)

Yeah, I agree.
Also the standard mention below in verification process. so we should definitely include those in the verification process.

> If a schema is provided in the Type Metadata, a recipient MUST validate the schema as defined in [Section 6.5](https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-09.html#schema-type-metadata).

I would suggest

- Let's just merge this pr with only type data and verification process would be following pr. if you're busy I could handle them. :)
- About checking schema and checking sd
  - include in the verification process, and when it failed throw an error. provide option to override this feature
  - include issuance process but default is not using. provide option to turn on

cre8:
I published a new version, where the types should be exported now

PR #289, opened 2025-06-19
Expose sd-jwt-vc public types

patatoid:


packages/sd-jwt-vc/package.json
```diff
@@ -9,6 +9,10 @@
     ".": {
       "import": "./dist/index.mjs",
       "require": "./dist/index.js"
+    },
+    "./types.js": {
```
patatoid, line 13: After tests the file does not look to be exposed. Any clues?

cre8:
You just need to add them to the `index.ts` file as an export, like:

```ts
export * from './sd-jwt-vc-config';
export * from './sd-jwt-vc-instance';
export * from './sd-jwt-vc-payload';
export * from './sd-jwt-vc-status-reference';
// new export
export * from './sd-jwt-vc-vct';
```

Then it will add the types to the. `index.d.ts` as an exportable one. You just need to add them there, no other changes needed

cre8:
closing this since it's solved in the other MR

PR #292, opened 2025-06-22
validate timestamp and set custom date

### LINKED ISSUE #291: Allow configuring `now` and `skewTime` for date based verifications
For example `iat` and `exp`, due to server skew it can be that verifications fail because they're just outside of the window.

We've seen status lists that are valid for only 15 seconds, often resulting in expired status lists checks. We could argue that 15 seconds validity time for a status list is very short (it is), but it would also be good to give some extra control to users of this library.

we usually allow providing:
- `now` - if you want to override the current time, instead of using `Date.now()`.
- `skewTime` - the time that should be taken as skew in verifications. If now = 100 and exp is 90, and skewTime is 20, then the verification will still succeed. I  think this should have a default value of higher than 0, probably a few seconds

See also https://curity.io/resources/learn/jwt-best-practices/

cre8:
closes #291 

Extending the validation of `iat`, `exp` and `nbf`. It is possible to pass a custom timestamp to the verify function.

We are not blocking credentials that will be issued with an `iat` value that is in the future, so the issuer has full control over it.

### DIFF
--- a/examples/sd-jwt-vc-example/all.ts
+++ b/examples/sd-jwt-vc-example/all.ts
@@ -56,7 +56,7 @@ import { createSignerVerifier, digest, ES256, generateSalt } from './utils';
   const credential = await sdjwt.issue(
     {
       iss: 'Issuer',
-      iat: new Date().getTime(),
+      iat: Math.floor(Date.now() / 1000),
       vct: 'ExampleCredentials',
       ...claims,
     },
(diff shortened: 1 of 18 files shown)


lukasjhan (approved):

TimoGlastra:
In my issue i also mentioned adding a skew time. What do you think about that @cre8?

Since setting current date doesn't work always (if the date is used for both exp and nbf).

cre8:
@TimoGlastra could you give an example where I need to validate with two different dates?

IMO I want to know for a specific timetamp (normally the current time) if a credential is valid. So I check if it is in the defined time range.

The skew time makes sense in use cases like time based token (where a token is valid for 15 secs or so). Then unsynced clocks become a problem and it would help for a better UX.


I cannot see any use case where somebody says (it's valid when it's 10 minutes earlier or 10 minutes later). But when you got the requirement from a use case, we can add it.

PR #293, opened 2025-06-23
Add-skew time

### LINKED ISSUE #291: Allow configuring `now` and `skewTime` for date based verifications

cre8:
closes #291

adding skew value in milliseconds. The default value is 0 since for some cases it is not required.

@TimoGlastra pls check if this is fine for you.

### DIFF
--- a/packages/core/src/jwt.ts
+++ b/packages/core/src/jwt.ts
@@ -16,14 +16,14 @@ export type JwtData<
  * Options for the JWT verifier
  */
 export type VerifierOptions = {
-/**
- * current time in seconds since epoch
- */
+  /**
+   * current time in seconds since epoch
+   */
   currentDate?: number;
 
- /**
-  * allowed skew for the current time in seconds. Positive value that will lower the iat and nbf checks, and increase the exp check.
-  */
+  /**
+   * allowed skew for the current time in seconds. Positive value that will lower the iat and nbf checks, and increase the exp check.
+   */
   skewSeconds?: number;
 };
 


packages/core/src/jwt.ts
```diff
@@ -117,21 +117,32 @@ export class Jwt<
...
    * @param verifier
-   * @param currentDate
+   * @param currentDate current time in seconds since epoch
+   * @param skew allowed skew in seconds for the current time in milliseconds. Positive value that will lower the iat and nbf checks, and increase the exp check.
    * @returns
    */
   public async verify(
     verifier: Verifier,
     currentDate = Math.floor(Date.now() / 1000),
+    skew: number = 0,
```
TimoGlastra, line 127: Would it make sense to:
- rename `skew` to `skewSeconds` (it can be hard to know in JS whether we're talking about seconds or milliseconds)
- make the `currentDate` and `skew` an object called `options`, this way adding new options will be simpler, and won't require a lot of positional arguments (which are harder to understand when just looking at the code).

```ts
verify(jwt, new Date(), 100)
```

vs

```ts
verify(jwt, {
  currentDate: new Date(), 
  skewSeconds: 100
})
```

cre8: You are right, I moved it to an option object and made both values in seconds.

packages/core/src/jwt.ts
```diff
@@ -12,6 +12,14 @@ export type JwtData<
   encoded?: string;
 };
 
+/**
+ * Options for the JWT verifier
+ */
+export type VerifierOptions = {
+  currentDate?: number; // current time in seconds since epoch
+  skewSeconds?: number; // allowed skew for the current time in seconds. Positive value that will lower the iat and nbf checks, and increase the exp check.
```
TimoGlastra, line 20: This way it will be visible as documentation in code editors.

```suggestion
/**
 * current time in seconds since epoch
 */
  currentDate?: number;

 /**
  * allowed skew for the current time in seconds. Positive value that will lower the iat and nbf checks, and increase the exp check.
  */
  skewSeconds?: number;
```

TimoGlastra (approved):
Nice!

I don't mind the breaking changes between 0.x versions to update the methods

cre8:
> Nice!
> 
> I don't mind the breaking changes between 0.x versions to update the methods

Okay, I will do another PR with the changes to keep this one smaller

Issue #288, opened 2025-06-19
Fetch VCT Metadata from SD-JWT header `vctm` claim


patatoid:
As stated in the [SD-JWT-based Verifiable Credentials (SD-JWT VC)](https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-09.html#name-from-type-metadata-glue-doc), VCT Metadata may be fetched from the SD-JWT header `vctm` claim formed by an array of encoded vct documents.

Getting vctm from the header should be compatible with the other ways to fetch vctm and can act as a primary source of information, using a remote registry being a second source in the case where no vctm is present in the header.

@gkatrakazas worked on this and may have insights about the implementation.

PR #294, opened 2025-06-24
feat: fetch VCT Metadata from SD JWT header (#288)

patatoid:
Referencing #288, here is suggested to take SD JWT header vctm claim as primary source of VCT Metadata if present.


packages/sd-jwt-vc/src/sd-jwt-vc-instance.ts
```diff
@@ -294,12 +294,50 @@ export class SDJwtVcInstance extends SDJwtInstance<SdJwtVcPayload> {
...
+      throw new Error('vctm claim in SD JWT header is invalid');
+    }
+
+    const typeMetadataFormat = (vctmHeader as unknown[])
+      .map((vctm) => {
+        if (!(typeof vctm === 'string')) {
+          throw new Error('vctm claim in SD JWT header is invalid');
+        }
+
+        return JSON.parse(Buffer.from(vctm, 'base64url').toString());
```
cre8, line 328: The buffer call will limit the usage to nodejs and exluce the browser. Pls use the `base64urlEncode` from the '@sd-jwt/utils' package

cre8 (approved):

PR #297, opened 2025-06-30
Add Verifier options

### LINKED ISSUE #296: Pass expected nonce to kb verification
The main goal is for the key binding to check if the key binding JWT includes the expected nonce that was passed before to the wallet e.g. via OID4VP.

It could be passed in the implementation of `kbVerifier`, but I doubt that all the nonces are knowns during init phase.

Instead I would extend the `verify` function like:

```typescript
    const kb = await sdjwt.kbJwt.verifyKB({
      verifier: this.userConfig.kbVerifier,
      payload: payload as JwtPayload,
      nonce,
    });
```

and change the verify function header like:

```typescript
  public async verify(
    encodedSDJwt: string,
    requiredClaimKeys?: string[],
    requireKeyBindingCode?: string,
  )
```

By changing the header of verify, we can also migrate to the VerifierOptions, so just encodedSDJwt is the first paramter and the rest is in the options as optional.

Also I do not know why the `validate` function is public. As a developer I would assume that there is one function to pass an sd jwt and I am able to configure the verification process.

cre8:
closes #296 

since we are adding more and more configurable parameters to the verify function, it was moved to an object.

This will cause a breaking change, but will make the changes easier in the future.

### DIFF
--- a/docs/0.x/keybinding.md
+++ b/docs/0.x/keybinding.md
@@ -40,6 +40,9 @@ const presentedSdJwt = await sdjwt.present(
 ## Verify
 
 ```ts
-const verified = await sdjwt.verify(presentedSdJwt, ['id', 'ssn'], true);
+const verified = await sdjwt.verify(presentedSdJwt, {
+  requiredClaimKeys: ['id', 'ssn'],
+  keyBindingNonce: '1234'
+});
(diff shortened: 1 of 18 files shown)


cre8:
> LGTM, If possible could you change the examples as well?

good point, I just checked that the test are running. I will update the examples ASAP.

PR #298, opened 2025-07-02
Use default fetcher to retrieve vct as URL

patatoid:
Here the point is to use the default fetcher to retrieve vct when it is expressed as an URL. The priority of vct retrieval would be then:
1. Use vctm in header if present
2. Use internal fetcher when vct is an URL
3. Use the custom (or default) fetcher for urns

### DIFF
--- a/packages/sd-jwt-vc/src/sd-jwt-vc-instance.ts
+++ b/packages/sd-jwt-vc/src/sd-jwt-vc-instance.ts
@@ -292,6 +292,10 @@ export class SDJwtVcInstance extends SDJwtInstance<SdJwtVcPayload> {
       return this.fetchVctFromHeader(result.payload.vct, result);
     }
 
+    if (result.payload.vct.startsWith('https')) {
+      return this.fetchVctFromUrl(result.payload.vct, result);
+    }
+
     const fetcher: VcTFetcher =
       this.userConfig.vctFetcher ??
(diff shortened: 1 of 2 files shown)


cre8:
Can you pls explain why you are skipping the fetcher in line 301? It is already covering your 2 and 3 case.

Even when the vct is a URL, I would maybe like to use my own fetcher (e.g. routing it via a proxy for privacy reasons).

cre8:
The default vctFetcher only handles resolving via url (because this is strait forward). But in case it's a uri, you of course need to pass your own fetcher that e.g. implements the connection to a database to look up the uri. Or in the case when you want to fetch it via a proxy.

As far as I understand the documentation, there is no need to call a fetch function to deal with the vctm and it is also not implemented [here](https://github.com/openwallet-foundation/sd-jwt-js/blob/121c6d9d3a9df09e9bbaf9a20b7759aa79ec091d/packages/sd-jwt-vc/src/sd-jwt-vc-instance.ts#L306)

Therefore I would close this PR

patatoid:
> But in case it's a uri, you of course need to pass your own fetcher that e.g. implements the connection to a database to look up the uri.

That's why I think the two, urn fetcher and url fetcher should be separated. That said, it is possible to have a switch case in the custom vctFetcher.

PR #303, opened 2025-08-01
fix: pass required options parameter

### LINKED ISSUE #302: Verify method on SdJwtVcInstance not utilizing VerifierOptions properties "skewSeconds" and "currentDate"
When calling the `verify` method on an SdJwtVcInstance object like the following:
```
const sdjwtvc = new SDJwtVcInstance({
    // verifier, hasher, etc...
});
const options: VerifierOptions = {
    keyBindingNonce,
    currentDate,
    skewSeconds,
};
await sdjwtvc.verify(credential, options);
```
It appears that the `currentDate` and `skewSeconds` properties from the `VerifierOptions` are ignored. Looking at the `verify` method in the parent `SdJwtInstance` class, I think the options are not being passed into the `validate` call. Below I have modified the code to pass it in:
```
  // This function is for verifying the SD JWT
  // If requiredClaimKeys is provided, it will check if the required claim keys are presentation in the SD JWT
  // If requireKeyBindings is true, it will check if the key binding JWT is presentation and verify it
  public async verify(encodedSDJwt: string, options?: VerifierOptions) {
    if (!this.userConfig.hasher) {
      throw new SDJWTException('Hasher not found');
    }
    const hasher = this.userConfig.hasher;

    const sdjwt = await SDJwt.fromEncode(encodedSDJwt, hasher);
    if (!sdjwt.jwt || !sdjwt.jwt.payload) {
      throw new SDJWTException('Invalid SD JWT');
    }
    const { payload, header } = await this.validate(encodedSDJwt, options); // <---- Note that this passing of `options` is missing in v0.14.0

   // Rest of `verify` logic follows...
```
Without this change I cannot specify a custom `currentDate` or `skewSeconds` when verifying an sd-jwt credential via an SdJwtVcInstance.

cre8:
closes #302

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -206,7 +206,7 @@ export class SDJwtInstance<ExtendedPayload extends SdJwtPayload> {
     if (!sdjwt.jwt || !sdjwt.jwt.payload) {
       throw new SDJWTException('Invalid SD JWT');
     }
-    const { payload, header } = await this.validate(encodedSDJwt);
+    const { payload, header } = await this.validate(encodedSDJwt, options);
 
     if (options?.requiredClaimKeys) {
       const keys = await sdjwt.keys(hasher);


lukasjhan (approved):
LGTM

PR #306, opened 2025-08-15
Fix/304

### LINKED ISSUE #304: Unable to verify an SD-JWT credential's status if the statuslist issuer is different from the credential's issuer
Assume we have an SD-JWT `myCredential` which contains a valid `status` reference.
When calling the `verify` method on `myCredential` via an `SdJwtVcInstance` object like the following:
```
const sdjwtvc = new SDJwtVcInstance({
    verifier: myVerifier,
    // hasher, etc...
});
await sdjwtvc.verify(credential);
```
It appears that the verifier `myVerifier` is used to verify both the signature of `myCredential` as well as the signature of the fetched statuslist credential. This works fine if the statuslist credential was issued by the same entity who issued `myCredential`, but not otherwise. My use-case involves having these roles be separate (which is also in the interest of mitigating the relevant privacy concerns defined in [Section 12.6 of the OAuth Token Statuslist Standard Draft 12](https://www.ietf.org/archive/id/draft-ietf-oauth-status-list-12.html#section-12.6)), but I can't seem to get this working via an `SdJwtVcInstance`.

I think the easiest change to allow this to work would be to introduce a new `SdJwtVcInstance` `userConfig` option along the lines of `statusListVerifier`, such that it can be used in this code snippet of `sd-jwt-vc-instance.ts` below. If not set, it could default to the `this.userConfig.verifier`, which is what is currently used.
```
  /**
   * Verifies the status of the SD-JWT-VC.
   * @param result
   * @param options
   */
  private async verifyStatus(
    result: VerificationResult,
    options?: VerifierOptions,
  ): Promise<void> {
 if (result.payload.status) {
      //checks if a status field is present in the payload based on https://www.ietf.org/archive/id/draft-ietf-oauth-status-list-02.html
      if (result.payload.status.status_list) {
        // fetch the status list from the uri
        const fetcher: StatusListFetcher =
          this.userConfig.statusListFetcher ??
          this.statusListFetcher.bind(this);
        // fetch the status list from the uri
        const statusListJWT = await fetcher(
          result.payload.status.status_list.uri,
        );

        const slJWT = Jwt.fromEncode<
          StatusListJWTHeaderParameters,
          StatusListJWTPayload
        >(statusListJWT);
        // check if the status list has a valid signature. The presence of the verifier is checked in the parent class.
        
        // *** Proposed changes here ***
        const statusListVerifier = this.userConfig.statusListVerifier ?? this.userConfig.verifier;
        await slJWT.verify(statusListVerifier as Verifier, options);

        // Rest of `verifyStatus` logic follows...
}
```

cre8:
Closes #304

Allows to pass a custom function for the status list JWT validation.

### DIFF
--- a/packages/sd-jwt-vc/README.md
+++ b/packages/sd-jwt-vc/README.md
@@ -87,6 +87,8 @@ Check out more details in our [documentation](https://github.com/openwallet-foun
 
 To add revocation capabilities, you can use the `@sd-jwt/jwt-status-list` library to create a JWT Status List and include it in the SD-JWT-VC.
 
+You can pass a dedicated `statusVerifier` function in the configuration to verify the signature of the payload of the JWT of the statuslist. If no function is provided, it will fallback to the verifier that is also used for the sd-jwt-vc.
+
 ### Type Metadata
 
 By setting the `loadTypeMetadataFormat` to `true` like this:
--- a/packages/sd-jwt-vc/src/sd-jwt-vc-config.ts
+++ b/packages/sd-jwt-vc/src/sd-jwt-vc-config.ts
@@ -1,4 +1,4 @@
-import type { SDJWTConfig } from '@sd-jwt/types';
+import type { SDJWTConfig, Verifier } from '@sd-jwt/types';
 import type { VcTFetcher } from './sd-jwt-vc-vct';
 
 export type StatusListFetcher = (uri: string) => Promise<string>;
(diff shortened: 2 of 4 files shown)


dalebowie:
If it's not too much trouble, would it be possible to please do a release with this change?

cre8:
@dalebowie sure, it was released in [0.15.0](https://github.com/openwallet-foundation/sd-jwt-js/releases/tag/v0.15.0)

PR #308, opened 2025-08-19
Fix/248

### LINKED ISSUE #248: Validate schema when issuing
Right now the schema behind vct is only validated when the credential is verified. To make sure only valid credentials are created, we also should validate the schema during issuance when there is a schema

cre8:
Closes #248

Adding validation during issuance when a schema is passed. The issue function will not fetch it from the vct, but requires direct passing.

Schema validation during verification was already included

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -97,7 +97,7 @@ export class SDJwtInstance<ExtendedPayload extends SdJwtPayload> {
     payload: Payload,
     disclosureFrame?: DisclosureFrame<Payload>,
     options?: {
-      header?: object; // This is for customizing the header of the jwt
+      header?: object; // This is for customizing the header of the jwt      
     },
   ): Promise<SDJWTCompact> {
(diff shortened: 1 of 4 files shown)


lukasjhan (approved):
LGTM

cre8:
There is an open topic in the sd jwt vc spec if the schema with required is used:
- to validate if the required fields are in the issued credentials
- to validate if the required fields are disclosed in the presented credential


The PR is right now covering the first part I think. Should we revert it and wait until we have a solution for it?

lukasjhan:
> There is an open topic in the sd jwt vc spec if the schema with required is used:
> 
>     * to validate if the required fields are in the issued credentials
> 
>     * to validate if the required fields are disclosed in the presented credential
> 
> 
> The PR is right now covering the first part I think. Should we revert it and wait until we have a solution for it?

They're discussing this issue? Hmm then I guess we should revert it and wait :)

lukasjhan:
@cre8 #312

PR #310, opened 2025-08-27
fix: make iss claim optional in SD-JWT VC payload

### LINKED ISSUE #309: iss claim should be optional
Hello,

I noticed that the `iss` claim is set as required.
https://github.com/openwallet-foundation/sd-jwt-js/blob/b8733bc33817900b2bfe946bd10e3ff8673bb5fc/packages/sd-jwt-vc/src/sd-jwt-vc-payload.ts#L5-L6
However, according to [RCF7519 4.1.1](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.1), the `iss` claim is OPTIONAL.
> [4.1.1](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.1).  "iss" (Issuer) Claim
>
>  The "iss" (issuer) claim identifies the principal that issued the
   JWT.  The processing of this claim is generally application specific.
   The "iss" value is a case-sensitive string containing a StringOrURI
   value.  Use of this claim is OPTIONAL.

Would you consider making this claim optional to align with the specification?

Thanks!

SakujiroInagaki:
closes #309


lukasjhan:
@SakujiroInagaki Could you sign off your commit?

SakujiroInagaki:
@lukasjhan I've added sign off. Could you please check? Thank you!

cre8 (approved):

PR #317, opened 2025-09-21
feat: add experimental vcld implementation

lukasjhan:


cre8:
@lukasjhan do you know if there is still a need for the vcld package? I haven't checked the spec if it is still aligned. But in general it would be fine for me to add it to this repo when the spec is stable about it.

PR #319, opened 2025-09-27
fix: add provenance to release step

### LINKED ISSUE #318: Migrate NPM workflows to Trusted Publishing
We need to migrate NPM releases to https://docs.npmjs.com/trusted-publishers and deprecate NPM Tokens.
Trusted publishing requires [npm CLI](https://docs.npmjs.com/cli/v11) version 11.5.1 or later.

We need to identify the dependencies and work needed to be done to upgrade to these versions.

Affected workflows:
./openwallet-foundation/sd-jwt-js/.github/workflows/release.yml
./openwallet-foundation/sd-jwt-js/.github/workflows/build-test-publish-on-push-cached.yaml

cre8:
fixes #318

Adding the flag accoding to: https://docs.npmjs.com/generating-provenance-statements#using-third-party-package-publishing-tools

Can not tested before merge, but should be correct. According to lerna this is already supported in v6, we have v8: https://github.com/lerna/lerna/issues/3657

### DIFF
--- a/.github/workflows/build-test-publish-on-push-cached.yaml
+++ b/.github/workflows/build-test-publish-on-push-cached.yaml
@@ -114,6 +114,7 @@ jobs:
     # needs permissions to write tags to the repository
     permissions:
       contents: write
+      id-token: write
     needs:
       - build
       - test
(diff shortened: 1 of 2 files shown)


lukasjhan:
@cre8 How about release 0.15.2 to test? It's also good for syncing the version.

cre8:
> @cre8 How about release 0.15.2 to test? It's also good for syncing the version.

Good idea, but I think for this we need to pass a config parameter to Lerna to force it. I will open a PR for this and then a new release

PR #320, opened 2025-10-01
fix: correct requiredClaimKeys in sd-jwt kb example to match presented claims

yunseorim1116:
## Summary
  Fix the SD-JWT key binding example to run without errors.

  ## Problem
  The example was failing with `SDJWTException: Missing required claim
   keys: ssn` because:
  - Only `id` was presented in the SD-JWT
  - But verification required both `ssn` and `id`

  ## Solution
  Changed `requiredClaimKeys` from `['ssn', 'id']` to `['id']` to match the presented claim.

Now the example runs successfully and demonstrates proper SD-JWT verification with key binding.

KakaoTalk_Photo_2025-10-01-12-22-52

### DIFF
--- a/examples/sd-jwt-example/kb.ts
+++ b/examples/sd-jwt-example/kb.ts
@@ -49,7 +49,7 @@ import { createSignerVerifier, digest, ES256, generateSalt } from './utils';
   );
 
   const verified = await sdjwt.verify(presentedSdJwt, {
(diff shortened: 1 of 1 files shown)


lukasjhan (approved):
Thank you for the contribution, @yunseorim1116  LGTM

PR #324, opened 2025-10-07
feat: updates all dependencies to latest versions + biome updates

cre8:
- pathing dependencies to latest version
- updating biome config files because of breaking changes (a lot of changes because of ordering of imports)


cre8:
Jose dropped support for version 18: https://github.com/panva/jose/releases/tag/v6.0.0

I would rather do the same than sticking to an older version, is this fine for you? @lukasjhan Then we would also increase the node version to 20 for this lib

lukasjhan (approved):
Good! Looks good to me

PR #325, opened 2025-10-07
fix: release ci, node version

cre8:
Adjust node version also in release github action, removing node version 18, adding 24


lukasjhan (approved):
LGTM

PR #332, opened 2025-11-13
Fix/sl-verification

### LINKED ISSUE #331: Different errors for status checks
Would be great to have different errors for JWT issues caused by statuslist checks.

Just spent days banging my head against the wall trying to figure out why a credential signature was invalid when in fact it was caused by the statuslist JWT signature.

cre8:
Fixes #331

- throws an error when status list validation fails. If a custom error message should be thrown, the verify funcation has to be overriden

### DIFF
--- a/packages/jwt-status-list/src/index.ts
+++ b/packages/jwt-status-list/src/index.ts
@@ -1,4 +1,4 @@
 export * from './status-list';
+export * from './status-list-exception';
 export * from './status-list-jwt';
 export * from './types';
-export * from './status-list-exception';
--- a/packages/jwt-status-list/src/status-list-jwt.ts
+++ b/packages/jwt-status-list/src/status-list-jwt.ts
@@ -1,13 +1,13 @@
 import type { JwtPayload } from '@sd-jwt/types';
 import { base64urlDecode } from '@sd-jwt/utils';
 import { StatusList } from './status-list';
+import { SLException } from './status-list-exception';
 import type {
   JWTwithStatusListPayload,
   StatusListEntry,
   StatusListJWTHeaderParameters,
   StatusListJWTPayload,
 } from './types';
-import { SLException } from './status-list-exception';
 
 /**
  * Decode a JWT and return the payload.
--- a/packages/jwt-status-list/src/status-list.ts
+++ b/packages/jwt-status-list/src/status-list.ts
@@ -1,7 +1,7 @@
 import { base64UrlToUint8Array, uint8ArrayToBase64Url } from '@sd-jwt/utils';
 import { deflate, inflate } from 'pako';
-import type { BitsPerStatus } from './types';
 import { SLException } from './status-list-exception';
+import type { BitsPerStatus } from './types';
 /**
  * StatusListManager is a class that manages a list of statuses with variable bit size.
  */


cre8:
@charsleysa please have a look if this approach is fine for you

packages/sd-jwt-vc/src/sd-jwt-vc-instance.ts
```diff
@@ -306,17 +306,47 @@ export class SDJwtVcInstance extends SDJwtInstance<SdJwtVcPayload> {
...
+        ) {
+          throw new SDJWTException('Verify Error: Status list is expired');
+        }
+        if (
+          slJWT.payload?.iat &&
+          (slJWT.payload.iat as number) - (options?.skewSeconds ?? 0) >
+            currentDate
+        ) {
+          throw new SDJWTException(
+            'Verify Error: Status list is not yet valid',
```
lukasjhan: Maybe we could mention which property throw an error. something like: "iat in Status list is not yet valid"

cre8:
Oh you are right (no work without coffee in the morning), I removed also the redundant exp check.

Instead of the first fix, I appended all error messages during the verify function so it is clear when the sl jwt fails, it is clear from the error message.

cre8:
> That would work. Though one downside is that it would suppress custom errors (you wouldn't be able to use instanceof checks).
> 
> Maybe it's better for the status list not to use the jwt verify at all and instead have it's own independent implementation.

this would end up in code duplication that I want to avoid. We may define a new Exception type like SLException?

charsleysa:
> this would end up in code duplication that I want to avoid. We may define a new Exception type like SLException?

That would still result in the suppression of custom errors from the user supplied verifier function. Though it does help in identifying exactly status list specific errors.

My thinking is the best way to avoid duplication would be to have check helper functions which return boolean results which can then be used from the status list verify function and it is the responsibility of the status list verify function to throw the errors.

lukasjhan (commented):
Looks good to me, Thanks.
Could you fix the lint error plz?

charsleysa:
> the sdjwtexception and slexception allow to pass a message and also an object. This object is passed so no information except the type of exception is lost

Ah, so the expectation is for users to throw SDJWTException or SLException instead of their own error class so that information can be correctly propagated?

charsleysa (approved):
Looks good!

packages/sd-jwt-vc/src/sd-jwt-vc-instance.ts
```diff
@@ -306,18 +307,18 @@ export class SDJwtVcInstance extends SDJwtInstance<SdJwtVcPayload> {
...
-        if (slJWT.payload?.exp && (slJWT.payload.exp as number) < currentDate) {
-          throw new SDJWTException('Status list is expired');
-        }
+        await slJWT
+          .verify(
+            this.userConfig.statusVerifier ??
+              (this.userConfig.verifier as Verifier),
+            options,
+          )
+          .catch((err: SLException) => {
```
TimoGlastra: I think casting this as an `SLException` is unsafe, since the method could throw other errors as well (since a custom `statusVerifier` can be passed).

PR #333, opened 2025-11-24
feat: add support for ES384 and ES512 in node-crypto and browser-crypto

dalebowie:


cre8:
Hey @dalebowie 
thank you for the contribution. I changed the class with just static functions to functions because it is not aligned with the biome ruleset.

cre8 (approved):
LGTM

PR #334, opened 2025-11-25
fix: status list error handling

cre8:
Improvement to https://github.com/openwallet-foundation/sd-jwt-js/pull/332

- rename exception to long form to make it clear
- if a custom error instance is thrown, it will be wrapped. if it's a status list exception, it will be passed

### DIFF
--- a/.github/workflows/build-test-publish-on-push-cached.yaml
+++ b/.github/workflows/build-test-publish-on-push-cached.yaml
@@ -105,8 +105,11 @@ jobs:
           node-version: 20
           cache: 'pnpm'
         # we are not using the github action for biome, but the package.json script. this makes sure we are using the same versions.
-      - name: Run Biome      
+      - name: Run Biome
         run: pnpm run biome:ci
+
+      - name: Run test typecheck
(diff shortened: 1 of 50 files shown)


charsleysa:
> * if a custom error instance is thrown, it will be passed. If not, it will be wrapped

By custom error instance, do you mean the StatusListException? If not, then the code is doing the opposite.

TimoGlastra (commented):
I think this is better. One improvement I'd suggest is to always keep the original error. A common pattern is to have a cause property on the error, so you can get a trace of wrapped errors. It can make debugging quite hard otherwise. 

You can implement the tostring method to include also the stack of the cause, so you see where the original error was thrown, and not just where it was wrapped.

TimoGlastra:
Erorr cause should be available in All modern browsers and Node.JS. I'm not sure about React Native.

We use a polyfill for it, but that was from 4 years ago, and it might not be needed anymore now.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause

PR #340, opened 2025-12-11
chore: some cleanup and typing improvements

TimoGlastra:
This cleans up some things:
- await promise expects (removes warnings from vitest)
- add typechecking for tests file (also in CI), there's some tests that haven't been updated based on the new API
- fix typo in `presenatedClaims`
- update biome commands, as the ci command was giving different results than lint/format locally


lukasjhan (approved):
Looks good. Thank you

PR #345, opened 2026-01-06
Fix/sl-iss

cre8:
According to the spec the `iss` is not a required field so it got removed from the required check: https://www.ietf.org/archive/id/draft-ietf-oauth-status-list-15.html#name-status-list-token-in-jwt-fo


lukasjhan (approved):
LGTM

PR #350, opened 2026-01-09
chore: merge releases into one file

TimoGlastra:
NPM Trusted Publishing only allows a single file, so i merged the stable and unstable releases files, similar to what we have in Credo

@cre8 can you check in the NPM page that all SD-JWT libraries have trusted publishing configured?

### DIFF
--- a/.github/workflows/build-test-publish-on-push-cached.yaml
+++ b/.github/workflows/build-test-publish-on-push-cached.yaml
@@ -1,4 +1,4 @@
-name: release
+name: build-test-publish
 on:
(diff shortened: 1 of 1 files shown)


cre8:
right now the other file got set as trusted so the latest release worked. Or do you think it's better to move release to a separate file instead of putting all in one?

TimoGlastra:
> Or do you think it's better to move release to a separate file instead of putting all in one?

It needs to be all in one, or at least all jobs that publish a package.

PR #357, opened 2026-01-13
fix: keep custom properties on type metadata

TimoGlastra:
We should retain custom properties on the type metadata. I don't think this should be treated as a breaking change, as it's a fix for the previous implementation


TimoGlastra:
If we can get this in a 0.18.1 that'd be great so I can update to 0.18 in Credo :)

cre8 (approved):

PR #358, opened 2026-01-16
chore: update to locale

TimoGlastra:
In draft 13 lang was changed to locale. I have implemented it right now that `lang` is automatically transformed to locale during validation. This ensures the output type always uses `locale` (new) but the input type can be both for now (to not break validation immediately).

Technically this is a breaking change, so not sure if we want to release this as 0.19 or 0.18.2?


cre8:
Ah, I didn't see the comment.

Pushing a feature to mark it as breaking should be fine for me

cre8 (approved):

PR #362, opened 2026-01-27
Fix/228

### LINKED ISSUE #228: Feat: return list of validations insteaf of throwing errors
@berendsliedrecht  and Timo came up with the wish that the verification functions will not throw errors and break up the function call, but rather return a list of validation with their result.

Before implementing this we have to make clear what are validations that can run without breaking up and what kind of response should they return. I want to avoid to introduce a complex validation framework for a scenario where the implementation it not worth the feature.

For me the validation of a signature is a must have. Is this wrong, we can not trust the JWT since it's not authentic! 
- expiration check
- not before check
- revoked check (checking the status list)

Could be done in parallel without stopping each other. But we have to make this clear to the implementer that just the function ran without throwing an error and being in the success call, the jwt does not have to be valid!

I also do not think that it is the job of the validation function to return some kind of validation report if everything is okay. This is out of scope of this library to generate reports that can be presented to the user and can be implemented by the ones needing the in their application.

We also have to think about if we want to pass a parameter like `throwErrorOnFail` to break up further checks. In case of efficiency this could be relevant.

cre8:
closes #228

adds safeverify that will run the checks in sequence without aborting at the first fail.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -10,6 +10,9 @@ import {
   type SDJWTCompact,
   type SDJWTConfig,
   type Signer,
+  type SafeVerifyResult,
+  type VerificationError,
+  type VerificationErrorCode,
 } from '@sd-jwt/types';
 import {
   base64urlDecode,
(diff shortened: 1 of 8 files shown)


cre8:
@TimoGlastra may you have a look at this since this was a feature requested by you :) It's not a breaking change since the default approach is not touched

lukasjhan (approved):
Looks good. 

I just like to make sure, let's discuss the interface before v1.0 :)

packages/core/src/index.ts
```diff
@@ -259,6 +262,202 @@ export class SDJwtInstance<ExtendedPayload extends SdJwtPayload, T = unknown> {
...
+    const addError = (
+      code: VerificationErrorCode,
+      message: string,
+      details?: unknown,
+    ) => {
+      errors.push({ code, message, details });
+    };
+
+    // Helper to convert exception to error code
+    const exceptionToCode = (error: Error): VerificationErrorCode => {
```
TimoGlastra, line 298: This feels a bit risky, what if we create a custom error class for each of these errors so that we can check with instanceof? Or we allow adding the code directly when the error is thrown, so that if the error is an sd jwt error, we just extract the code

TimoGlastra: Can even remap error to different code on a higher level (e.g. signature verification failed can be mapped to status verification failed when verifying a status list)

cre8: an alternative would be instead of adding keys, we just return a list of errors. In this case we drop the need of a mapping and the user can device if they want to use a pipe or so to transform the errors before e.g. presenting it to the user. WDYT?

TimoGlastra: I'd prefer custom error codes, as that makes it more explicit what type of error can be thrown / should be handled, and also gives some type hinting. Since you already set everything up for error codes I think it makes sense to keep it, but it's also something we can add in a follow up PR without breaking changes

PR #364, opened 2026-01-28
feat: allow status list to be issued as cwt

cre8:
Allows to issue the status list also as CWT compared to JWT.

Also adds some reusable variables that are defined in the spec.

A renaming of the package is not done. When this gets moved to the new Lab as a core library, new package names can be defined (core for bit string logic, one for JWT and one for CWT to minimize dependencies in them)


TimoGlastra (approved):
I would like to merge this with our generic COSE/CWT/CBOR signing/verification we have in our mdoc lib and move that to the common library, but for now this is great!

PR #378, opened 2026-06-23
fix: verify common jwt claims for kb-jwt

### LINKED ISSUE #377: KB JWT is not verified against standard JWT requirements
E.g. the `iat` / `exp` values in the KB-JWT are not verified. Found this while doing OIDC conformance testing.  Would it make sense for the KB-JWT verifier to also run the base JWT verification logic?

cc @cre8

TimoGlastra:
Fixes #377

### DIFF
--- a/packages/core/src/kbjwt.ts
+++ b/packages/core/src/kbjwt.ts
@@ -50,8 +50,8 @@ export class KBJwt<
 
     // Delegate signature verification and common JWT claim validation
     // (iat, nbf, exp) to the shared Jwt.verify implementation. The kbVerifier
-    // needs the kb+jwt payload (e.g. the holder's cnf key) which the base
-    // verifier receives via the options argument.
+    // needs the kb+jwt payload (e.g. the holder's cnf key), so we wrap it to
+    // forward values.payload instead of the base verifier's options argument.
     await this.verify(
       (data, sig) => values.verifier(data, sig, values.payload),
       values.options,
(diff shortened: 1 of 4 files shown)


packages/core/src/kbjwt.ts
```diff
@@ -39,19 +44,19 @@ export class KBJwt<
...
-      values.payload,
-    );
-    if (!verified) {
-      throw new SDJWTException('Verify Error: Invalid JWT Signature');
-    }
     if (this.payload.nonce !== values.nonce) {
       throw new SDJWTException('Verify Error: Invalid Nonce');
     }
 
+    // Delegate signature verification and common JWT claim validation
```
berendsliedrecht, line 51: If the spec mandates that some values are required, don't we want to check that instead of the letting the user check it? Or would that complicate the PR a lot?

TimoGlastra: They are not required, but if present we want to check them.

The OID4VP HAIP conformance test suite checks that we reject KB-JWTs where the `iat` is in the future (and past, but that's maybe harder to solve in this library, what is "past"? it makes sense to reject an iat of one year ago, but there's no strict timeline).

TimoGlastra: `iat` is required, and that is check. But it's not checked against the current time. so it can be any value as long as it's defined. This verifies that the `iat` is not in the future (and will add a check to credo to make it reasonably in the current time (e.g. created X minutes ago). But that feels too opiniated for this library

berendsliedrecht (approved):

cre8 (approved):

berendsliedrecht (approved):

PR #389, opened 2026-07-18
Fix RFC 9901 audit findings

### LINKED ISSUE #388: conformance audit to rfc 9901
# RFC 9901 Conformance Audit of `@sd-jwt/core`

**Project:** `openwallet-foundation/sd-jwt-js`  
**Package:** `packages/core` (`@sd-jwt/core` 0.20.0)  
**Audited revision:** [`c95e1abfeb39c2814b874396cb03f1b7ef5d478c`](https://github.com/openwallet-foundation/sd-jwt-js/commit/c95e1abfeb39c2814b874396cb03f1b7ef5d478c) (`main`, 2026-07-03)  
**Audit date:** 2026-07-18  
**Primary standard:** [RFC 9901 — Selective Disclosure for JSON Web Tokens](https://www.rfc-editor.org/rfc/rfc9901.html)  
**Normative dependencies considered:** [RFC 7515 (JWS)](https://www.rfc-editor.org/rfc/rfc7515.html), [RFC 7519 (JWT)](https://www.rfc-editor.org/rfc/rfc7519.html), [RFC 8725 (JWT BCP)](https://www.rfc-editor.org/rfc/rfc8725.html), [RFC 7800 (`cnf`)](https://www.rfc-editor.org/rfc/rfc7800.html), and the [IANA Named Information Hash Algorithm Registry](https://www.iana.org/assignments/named-information/named-information.xhtml).

## Executive conclusion

`@sd-jwt/core` implements the central SD-JWT mechanisms correctly on the happy path: object and array Disclosures, recursive disclosure processing, digest calculation, decoys, compact encoding, presentation selection, Key Binding `sd_hash`, and both JWS JSON serializations are present. The recently added RFC 9901 checks for duplicate digest values, unreferenced Disclosures, and collisions with permanently disclosed claims are valuable.

However, the package should **not yet be released as 1.0 with an RFC 9901-conformance claim**. The verifier accepts several inputs that RFC 9901 requires it to reject. The most security-relevant gaps are:

1. malformed `_sd`, `...`, and Disclosure structures are accepted;
2. validity claims are checked before Disclosure processing, so a disclosed expired `exp` can be accepted;
3. KB-JWT audience and maximum age are not validated;
4. an Issuer-signed JWT with `alg: none` can be accepted;
5. hash algorithms explicitly described by RFC 9901 as unfit for security applications are accepted;
6. a Holder presentation can accept and preserve an incoming KB-JWT instead of rejecting it;
7. salt uniqueness is not enforced;
8. General JSON parsing does not enforce the first-unprotected-header rule and its verifier skips JWT time validation;
9. dot-delimited presentation paths are ambiguous for valid JSON claim names and can select the wrong Disclosure.

The audit reproduced **24 negative cases that the current implementation accepts**. These are not theoretical API observations; each was confirmed in an executable test harness against the audited revision.

## Scope and method

The review covered every RFC 9901 requirement that can reasonably be implemented or enforced by a generic core library. Requirements belonging to a deployment, application profile, transport, storage system, or key-management system are marked as delegated rather than treated as source-code defects.

The audit used:

- line-by-line review of `packages/core/src` and all package tests;
- mapping of RFC 9901 normative language to issuance, Holder, and Verifier behavior;
- official IANA registry comparison;
- package build and test execution;
- adversarial tests for malformed serialization, malformed Disclosures, validation ordering, Key Binding, salt reuse, JSON serialization, and presentation selection.

## Verification results

| Check                                                         |                                                        Result |
| ------------------------------------------------------------- | ------------------------------------------------------------: |
| Node unit-test selection used by `test:node`                  |                                                     81 passed |
| jsdom selection used by `test:browser`                        |                                                     81 passed |
| Full Node test discovery, including nested and `test/` suites |                                                    159 passed |
| TypeScript/CJS/ESM/declaration build                          |                                                        Passed |
| Statement coverage                                            |                                                        75.17% |
| Branch coverage                                               |                                                        69.88% |
| Line coverage                                                 |                                                        75.98% |
| `src/index.ts` line coverage                                  |                                                        44.47% |
| Additional negative audit probes                              | 24/24 confirmed current acceptance of invalid or unsafe cases |

The package's normal tests passing does not contradict the findings: most gaps are missing rejection tests. The existing RFC-specific suite covers three Section 7.1 cases but not the complete validation algorithm.

### C-01 — Malformed SD-JWT structures are accepted

**Severity:** Critical  
**RFC:** [Sections 4.1, 4.2.1, 4.2.2, 4.2.4, and 7.1](https://www.rfc-editor.org/rfc/rfc9901.html#section-7.1)

The processing code detects digest values but does not strictly validate the containing structures. Confirmed accepted cases include:

- `_sd` with a non-array value;
- an object `_sd` referencing a two-element array Disclosure;
- an array placeholder referencing a three-element object-property Disclosure;
- an array `{"...": digest}` placeholder with additional members;
- non-string salt and claim-name values;
- `_sd` and `...` as disclosed claim names;
- nested `_sd_alg`;
- two Disclosures producing the same claim name at one object level, with the latter silently overwriting the former.

The main causes are permissive runtime construction in [`Disclosure`](https://github.com/openwallet-foundation/sd-jwt-js/blob/c95e1abfeb39c2814b874396cb03f1b7ef5d478c/packages/core/src/utils/disclosure.ts#L13-L32) and shape-insensitive processing in [`unpackArray` and `unpackObjInternal`](https://github.com/openwallet-foundation/sd-jwt-js/blob/c95e1abfeb39c2814b874396cb03f1b7ef5d478c/packages/core/src/decode/decode.ts#L140-L271).

**Required change:** introduce one strict recursive validation/processing algorithm matching RFC 9901 Section 7.1 exactly. Validate runtime JSON types and arity before inserting any claim. Reject every malformed reserved structure rather than treating it as ordinary JSON or ignoring it.

### C-02 — Validity claims are validated before Disclosure processing

**Severity:** Critical  
**RFC:** [Section 7.1, step 6](https://www.rfc-editor.org/rfc/rfc9901.html#section-7.1), [Section 9.7](https://www.rfc-editor.org/rfc/rfc9901.html#section-9.7), and [RFC 7519 Section 4.1.4](https://www.rfc-editor.org/rfc/rfc7519.html#section-4.1.4)

[`validate()`](https://github.com/openwallet-foundation/sd-jwt-js/blob/c95e1abfeb39c2814b874396cb03f1b7ef5d478c/packages/core/src/index.ts#L508-L525) calls `Jwt.verify()` on the raw Issuer-signed payload and only afterwards reconstructs the Processed SD-JWT Payload. Consequently, an `exp`, `nbf`, or `aud` supplied through a Disclosure is not validated. A credential with disclosed `exp: 1` was accepted at `currentDate: 100`.

Additional JWT validation defects:

- `aud` is not validated by the core verifier;
- non-numeric `exp`, `nbf`, or `iat` values are ignored rather than rejected;
- `exp === currentDate` is accepted because the comparison uses `<`, although the current time must be before `exp`;
- the General JSON verifier performs no `exp`, `nbf`, or `aud` validation.

**Required change:** first verify the JWS signature, then process and strictly validate Disclosures, then validate required and validity-controlling claims on the Processed SD-JWT Payload. Add expected-audience support and strict NumericDate checks.

### C-03 — Key Binding does not validate audience or an acceptable `iat` window

**Severity:** Critical  
**RFC:** [Sections 4.3 and 7.3](https://www.rfc-editor.org/rfc/rfc9901.html#section-4.3)

The KB-JWT implementation correctly emits `typ: kb+jwt`, includes `sd_hash`, checks the nonce, verifies a callback signature, and compares the recomputed `sd_hash`. However:

- there is no expected KB audience option and no comparison of `aud` to the current Verifier;
- `iat` is only rejected when it is in the future; no maximum age or acceptable time window is enforced;
- runtime types for `iat`, `aud`, `nonce`, and `sd_hash` are not strictly checked;
- a string `iat` is accepted;
- the public `KBJwt.verifyKB()` retains draft-era `_sd_hash` acceptance.

See [`VerifierOptions`](https://github.com/openwallet-foundation/sd-jwt-js/blob/c95e1abfeb39c2814b874396cb03f1b7ef5d478c/packages/core/src/jwt.ts#L18-L50) and [`KBJwt.verifyKB()`](https://github.com/openwallet-foundation/sd-jwt-js/blob/c95e1abfeb39c2814b874396cb03f1b7ef5d478c/packages/core/src/kbjwt.ts#L12-L60).

**Required change:** add explicit `expectedKeyBindingAudience`, `keyBindingMaxAgeSeconds` (or an accepted time interval), and strict runtime claim validation. Keep clock skew separate from maximum age.

### C-04 — `alg: none` and unsafe signature-algorithm policy are not rejected centrally

**Severity:** Critical  
**RFC:** [Sections 4.1, 7.1, and 9.1](https://www.rfc-editor.org/rfc/rfc9901.html#section-9.1), plus [RFC 8725 Sections 3.1 and 3.2](https://www.rfc-editor.org/rfc/rfc8725.html#section-3.1)

`Jwt.verify()` never checks the JOSE `alg` header. If the injected verifier returns `true`, an Issuer-signed JWT with `alg: none` is accepted. Issuance likewise accepts `signAlg: "none"`. See [`Jwt.verify()`](https://github.com/openwallet-foundation/sd-jwt-js/blob/c95e1abfeb39c2814b874396cb03f1b7ef5d478c/packages/core/src/jwt.ts#L161-L188) and [`SDJwtInstance.issue()`](https://github.com/openwallet-foundation/sd-jwt-js/blob/c95e1abfeb39c2814b874396cb03f1b7ef5d478c/packages/core/src/index.ts#L121-L150).

The signature callback remains the right abstraction for cryptographic verification and Issuer-key resolution, but the core parser must unconditionally reject `none` and should enforce an application-configured allowlist before invoking the callback.

### C-05 — Insecure truncated hash algorithms are accepted

**Severity:** Critical  
**RFC:** [Sections 4.1.1 and 9.4](https://www.rfc-editor.org/rfc/rfc9901.html#section-9.4)

The type and constructor allow every entry in the IANA registry, including `sha-256-32`, `sha-256-64`, `sha-256-96`, and other truncated variants. RFC 9901 explicitly warns that registry presence alone does not establish suitability and names `sha-256-32` and `sha-256-64` as unfit for security applications.

The current list is visible in [`types/type.ts`](https://github.com/openwallet-foundation/sd-jwt-js/blob/c95e1abfeb39c2814b874396cb03f1b7ef5d478c/packages/core/src/types/type.ts#L12-L31), while acceptance is controlled in [`SDJwtInstance`](https://github.com/openwallet-foundation/sd-jwt-js/blob/c95e1abfeb39c2814b874396cb03f1b7ef5d478c/packages/core/src/index.ts#L53-L64) and [`getSDAlgAndPayload`](https://github.com/openwallet-foundation/sd-jwt-js/blob/c95e1abfeb39c2814b874396cb03f1b7ef5d478c/packages/core/src/decode/decode.ts#L303-L321).

**Required change:** distinguish registry identifiers from algorithms allowed by security policy. Default to a conservative secure allowlist and permit profiles to narrow or explicitly extend it. A present but non-string `_sd_alg` must be rejected rather than treated as absent.

### H-01 — Compact parsing accepts a bare JWT as an SD-JWT

**Severity:** High  
**RFC:** [Section 4](https://www.rfc-editor.org/rfc/rfc9901.html#section-4)

RFC 9901 requires a regular SD-JWT to end in `~`. The compatibility branches in [`splitSdJwt()` and `decodeSdJwt()`](https://github.com/openwallet-foundation/sd-jwt-js/blob/c95e1abfeb39c2814b874396cb03f1b7ef5d478c/packages/core/src/decode/decode.ts#L34-L92) accept a compact JWT with no tilde. The high-level verifier therefore cannot reliably distinguish a JWT from an SD-JWT.

**Required change:** make strict RFC 9901 parsing the default for 1.0. If plain-JWT compatibility is still needed, expose it through a separately named opt-in API.

### H-02 — Holder processing accepts an incoming SD-JWT+KB

**Severity:** High  
**RFC:** [Section 7.2](https://www.rfc-editor.org/rfc/rfc9901.html#section-7.2)

RFC 9901 requires the Holder to reject an SD-JWT+KB received at issuance. `present()` accepts one and, when no new KB options are supplied, preserves the existing KB-JWT through [`SDJwt.present()`](https://github.com/openwallet-foundation/sd-jwt-js/blob/c95e1abfeb39c2814b874396cb03f1b7ef5d478c/packages/core/src/sdjwt.ts#L124-L137) and [`SDJwtInstance.present()`](https://github.com/openwallet-foundation/sd-jwt-js/blob/c95e1abfeb39c2814b874396cb03f1b7ef5d478c/packages/core/src/index.ts#L194-L225).

The Holder API also does not itself verify the Issuer signature before presenting; callers must remember to run a separate verification workflow.

**Required change:** provide an explicit Holder acceptance function that performs Section 7.1 and rejects any incoming KB-JWT. Presentation should only operate on a validated Holder object or perform this check itself.

### H-03 — Salt requirements are only requested, not enforced

**Severity:** High  
**RFC:** [Sections 4.2.1 and 9.3](https://www.rfc-editor.org/rfc/rfc9901.html#section-9.3)

The packer requests `saltGenerator(16)`, but it does not validate that the result is a string, contains at least 128 bits of cryptographically secure randomness, or is unique across all claims in the credential. A deterministic generator returning the same value produced two Disclosures with identical salts.

See [`pack()`](https://github.com/openwallet-foundation/sd-jwt-js/blob/c95e1abfeb39c2814b874396cb03f1b7ef5d478c/packages/core/src/sdjwt.ts#L260-L356).

**Required change:** define the generator contract unambiguously in bytes/bits, provide a secure built-in default where possible, validate the returned runtime type and minimum encoded entropy, and reject duplicate salts/digests within one issuance operation. Cross-credential uniqueness/randomness remains the generator's responsibility and must be documented.

### H-04 — General JSON parsing and verification are incomplete

**Severity:** High  
**RFC:** [Section 8](https://www.rfc-editor.org/rfc/rfc9901.html#section-8)

Serialization places `disclosures` and `kb_jwt` only in the first unprotected header, but `GeneralJSON.fromSerialized()` does not reject those members in later unprotected headers; it silently discards them. It also discards arbitrary unprotected header parameters instead of preserving them. See [`GeneralJSON.fromSerialized()`](https://github.com/openwallet-foundation/sd-jwt-js/blob/c95e1abfeb39c2814b874396cb03f1b7ef5d478c/packages/core/src/generalJSON.ts#L67-L84).

`SDJwtGeneralJSONInstance.validate()` verifies signatures and reconstructs claims but does not apply the JWT validity checks used by the compact verifier. See [`index.ts`](https://github.com/openwallet-foundation/sd-jwt-js/blob/c95e1abfeb39c2814b874396cb03f1b7ef5d478c/packages/core/src/index.ts#L868-L911).

### H-05 — Dot-delimited presentation paths are ambiguous

**Severity:** High (privacy impact)  
**RFC:** [Sections 3.2 and 7.2](https://www.rfc-editor.org/rfc/rfc9901.html#section-7.2)

Claim names are arbitrary JSON strings, including strings containing `.` or numeric-looking names. Internally, the library flattens presentation paths to strings such as `a.b`. A top-level claim named `a.b` collides with nested claim `a -> b`. In a reproduced case, requesting the nested claim selected the top-level Disclosure instead.

See [`transformPresentationFrame()`](https://github.com/openwallet-foundation/sd-jwt-js/blob/c95e1abfeb39c2814b874396cb03f1b7ef5d478c/packages/core/src/present/present.ts#L102-L124) and construction of `disclosureKeymap` in [`decode.ts`](https://github.com/openwallet-foundation/sd-jwt-js/blob/c95e1abfeb39c2814b874396cb03f1b7ef5d478c/packages/core/src/decode/decode.ts#L222-L265).

**Required change:** use an unambiguous path representation, preferably an array of path segments or properly escaped JSON Pointer. This is best fixed before 1.0 because it changes public presentation and required-claim APIs.

### M-01 — Key-to-Holder and Issuer-to-key checks are delegated but under-specified

**Severity:** Medium  
**RFC:** [Sections 4.3.2 and 7.3](https://www.rfc-editor.org/rfc/rfc9901.html#section-7.3)

The injected issuer and KB verifier callbacks can implement the required trust decisions. The core library passes the Processed SD-JWT Payload to the KB verifier, which is useful for resolving `cnf`. However, the library cannot ensure that callbacks actually validate the Issuer/key relationship or use the Holder key identified by `cnf`.

This is acceptable as an abstraction boundary only if the API contract is explicit and verified by integration tests. The callback should receive parsed protected-header data and an explicit verification context, and documentation should state every required check that remains the callback's responsibility.

### M-02 — Test coverage is not sufficient for a stable security API

**Severity:** Medium

The complete suite passes, but the orchestration class holding issuance, presentation, verification, and JSON serialization behavior has only 44.47% line coverage. There is no comprehensive official RFC 9901 Appendix-vector test suite, and the package's `test:node`/`test:browser` glob only selects top-level `src/test/*.spec.ts`; nested tests are reached later through the coverage command but not in those two phases.

Before 1.0, add:

- all applicable RFC 9901 Appendix vectors as fixed interoperability tests;
- strict negative tests for every Section 7.1 rejection branch;
- role/workflow tests for Section 7.2 and Section 7.3;
- property-based or fuzz tests for JSON trees, reserved names, arrays, recursion, malformed compact inputs, and duplicate values;
- differential interoperability tests against at least one independent RFC 9901 implementation.

### M-03 — Release metadata still carries pre-RFC wording

**Severity:** Medium/low

`packages/core/package.json` still describes the package as an “sd-jwt draft 7 implementation,” while the README refers to RFC 9901. This should be corrected together with an explicit conformance/support statement, the callback responsibility matrix, supported algorithms, and migration notes for strict 1.0 parsing.

## Requirement-by-requirement matrix

Legend: **Conforms** = implemented for the audited scope; **Partial** = useful implementation exists but a normative case is missing; **Fails** = confirmed non-conforming behavior; **Delegated** = must be guaranteed by the application/profile/callback and should be documented; **Out of scope** = not reasonably enforceable in this package.

### Format, issuance, and Disclosure construction

| RFC 9901 requirement                                              | Status                           | Evidence / assessment                                                                                                           |
| ----------------------------------------------------------------- | -------------------------------- | ------------------------------------------------------------------------------------------------------------------------------- |
| §4 compact component ordering                                     | Conforms for encoding            | `encodeSDJwt()` emits Issuer JWT, Disclosures, final empty value or KB-JWT.                                                     |
| §4 trailing `~` for SD-JWT without KB                             | **Fails on parsing**             | Encoder is correct; decoder accepts a bare JWT.                                                                                 |
| §4 Holder sends no unknown or duplicate Disclosure                | **Partial**                      | Presentation selection uses issued Disclosures, but duplicate encoded Disclosures supplied as input are accepted.               |
| §4.1 Issuer-signed JWT is signed                                  | Partial/delegated                | A signer is required for issuance and a verifier callback for verification. Issuer trust/key ownership remains delegated.       |
| §4.1 `alg` is not `none`                                          | **Fails**                        | Issuance and verification accept `none` if callbacks do.                                                                        |
| §4.1 payload is a JSON object                                     | **Partial**                      | TypeScript types say object; runtime decoding does not consistently enforce it.                                                 |
| §4.1 `_sd` and `...` appear only as digest containers             | **Fails**                        | Issuer validation omits `...`; verifier accepts multiple malformed uses.                                                        |
| §4.1 same digest never appears twice                              | Partial                          | Verifier rejects repeated embedded digest values; issuer does not guard against generator/decoy collisions.                     |
| §4.1 explicit application typing                                  | Delegated                        | Supported through headers/subclasses, but not a generic core default. Profile must set and verify it.                           |
| §4.1.1 `_sd_alg` only at top level                                | **Fails on verification**        | Issuance rejects nested `_sd_alg`; verifier accepts it.                                                                         |
| §4.1.1 absent `_sd_alg` defaults to `sha-256`                     | Conforms                         | Implemented.                                                                                                                    |
| §4.1.1 `_sd_alg` is a registered, case-sensitive string           | **Partial**                      | Registry list is current; a present non-string value silently defaults to SHA-256.                                              |
| §4.1.1 SHA-256 support                                            | Conforms through injected hasher | SHA-256 is the default; cryptographic implementation is injected.                                                               |
| §4.1.2 Holder key/reference via `cnf` for Key Binding             | Delegated                        | Payload supports `cnf`; application controls issuance and key resolution.                                                       |
| §4.2.1 object Disclosure has three elements                       | **Partial**                      | Issuance normally creates three; verification accepts two-element form at `_sd`.                                                |
| §4.2.1 salt and claim name are strings                            | **Fails at runtime validation**  | Non-string values are accepted from decoded input.                                                                              |
| §4.2.1 salt unique and secret                                     | **Partial/delegated**            | No duplicate check; confidentiality is outside package scope.                                                                   |
| §4.2.1 claim name is not `_sd`, `...`, or an existing clear claim | **Partial**                      | Collision with a clear claim is checked; reserved names and collisions between two Disclosures are not fully rejected.          |
| §4.2.1 UTF-8 JSON then unpadded base64url                         | Conforms                         | Strict UTF-8/JSON decoding and base64url utility are used.                                                                      |
| §4.2.2 array Disclosure has exactly two elements                  | **Partial**                      | Issuance normally creates two; verifier accepts three at an array placeholder.                                                  |
| §4.2.3 digest input is ASCII Disclosure string; output base64url  | Conforms                         | Digest is computed over the original encoded Disclosure string.                                                                 |
| §4.2.4.1 `_sd` is an array of strings                             | **Fails on verification**        | Non-array and invalid element types are not strictly rejected.                                                                  |
| §4.2.4.1 hide original object-property order                      | Conforms                         | `_sd` digests are sorted.                                                                                                       |
| §4.2.4.2 array placeholder has exactly one `...` string member    | **Fails on verification**        | Extra members and wrong Disclosure arity are accepted.                                                                          |
| §4.2.5 decoy hash uses the same algorithm and base64url           | Conforms                         | Implemented; randomness quality is delegated to the generator.                                                                  |
| §4.2.6 recursive Disclosure dependencies                          | Conforms                         | Recursive processing plus unreferenced-Disclosure rejection enforces connection; presentation-frame traversal includes parents. |

### Key Binding

| RFC 9901 requirement                                                 | Status                                     | Evidence / assessment                                                                                             |
| -------------------------------------------------------------------- | ------------------------------------------ | ----------------------------------------------------------------------------------------------------------------- |
| §4.3 `typ: kb+jwt`, signed `alg`, no `none`                          | Conforms for normal high-level flow        | Created correctly and checked; callback enforces cryptographic algorithm/key details.                             |
| §4.3 required `iat`, single-string `aud`, string `nonce`, `sd_hash`  | **Partial**                                | Presence is checked using truthiness; strict runtime types are not.                                               |
| §4.3.1 `sd_hash` over exact presented SD-JWT and same hash algorithm | Conforms                                   | Recomputed and compared in high-level verification.                                                               |
| §4.3.2 KB signature uses Holder key from SD-JWT                      | Delegated                                  | The callback receives processed payload but core cannot prove which key callback used.                            |
| §7.3 KB-required decision independent of KB presence                 | Conforms at API-policy level               | Caller opts in through a nonce; missing KB is then rejected. A clearer `requireKeyBinding` option is recommended. |
| §7.3 validate KB signature and `typ`                                 | Conforms through callback plus core checks | `typ` checked; signature callback result must be true.                                                            |
| §7.3 validate `nonce` and audience                                   | **Partial**                                | Nonce compared; expected audience not supported.                                                                  |
| §7.3 `iat` in acceptable window                                      | **Fails**                                  | Only future time is rejected; stale KB-JWT is accepted.                                                           |
| §7.3 KB-JWT valid in all other JWT respects                          | **Partial**                                | NumericDate/type/audience gaps remain.                                                                            |

### Verification and Holder processing

| RFC 9901 requirement                                            | Status                         | Evidence / assessment                                                                              |
| --------------------------------------------------------------- | ------------------------------ | -------------------------------------------------------------------------------------------------- |
| §7.1 split SD-JWT into required components                      | **Partial**                    | Standard inputs split correctly; bare JWT compatibility violates strict grammar.                   |
| §7.1 validate secure Issuer algorithm and reject `none`         | **Fails**                      | No central check or allowlist.                                                                     |
| §7.1 validate signature, Issuer, and Issuer key                 | Delegated                      | Verifier callback must perform all trust resolution.                                               |
| §7.1 understand and approve `_sd_alg`                           | **Partial**                    | Registry membership checked, security suitability not checked.                                     |
| §7.1 calculate each Disclosure digest                           | Conforms                       | Implemented using original encoded bytes/string.                                                   |
| §7.1 identify only valid `_sd` and `...` digest locations       | **Fails**                      | Structure validation is incomplete.                                                                |
| §7.1 reject wrong Disclosure arity by digest location           | **Fails**                      | Confirmed both directions.                                                                         |
| §7.1 reject reserved/colliding claim names                      | **Partial**                    | Cleartext collision passes; reserved and disclosure-to-disclosure collisions fail.                 |
| §7.1 ignore unmatched digests / remove unmatched array elements | Conforms for valid structures  | Implemented.                                                                                       |
| §7.1 remove `_sd` and top-level `_sd_alg`                       | Conforms for valid structures  | Implemented; malformed/nested `_sd_alg` caveat applies.                                            |
| §7.1 reject duplicate embedded digests                          | Conforms                       | Implemented recursively.                                                                           |
| §7.1 reject every unreferenced Disclosure                       | **Partial**                    | Unique unreferenced digest rejected; duplicate copies collapse in the digest map and are accepted. |
| §7.1 validate required and validity claims on processed payload | **Fails/partial**              | Required keys supported; actual temporal/audience validation occurs on raw payload.                |
| §7.1 abort on any failed step                                   | **Partial**                    | Error propagation works, but several required failure conditions are not detected.                 |
| §7.2 Holder rejects received SD-JWT+KB                          | **Fails**                      | Presentation accepts and preserves it.                                                             |
| §7.2 Holder validates received SD-JWT                           | Delegated/partial              | `present()` processes Disclosures but does not itself verify Issuer signature.                     |
| §7.2 selected Disclosures connect to signed payload             | Conforms for unambiguous paths | Parent Disclosures are selected; dot-path ambiguity remains.                                       |

### JWS JSON Serialization and security considerations

| RFC 9901 requirement                                                       | Status                          | Evidence / assessment                                                                                       |
| -------------------------------------------------------------------------- | ------------------------------- | ----------------------------------------------------------------------------------------------------------- |
| §8 Flattened JSON Serialization                                            | Conforms for normal round trips | Implemented. Strict input schema checks should be expanded.                                                 |
| §8 General JSON Serialization                                              | Partial                         | Normal round trips work; later-header prohibition and header preservation do not.                           |
| §8 `kb_jwt` present for JSON SD-JWT+KB and compact form used for `sd_hash` | Conforms in high-level flow     | Temporary compact representation is used.                                                                   |
| §8.3 `disclosures`/`kb_jwt` only in first unprotected header               | **Fails on parsing**            | Serialization is correct; parser silently ignores forbidden later occurrences.                              |
| §9.1 always verify Issuer signature                                        | Partial/delegated               | High-level verifier calls the callback; Holder `present()` can be used without prior verification.          |
| §9.3 cryptographic, independent, ≥128-bit salts                            | **Partial/delegated**           | `16` is requested; quality, length interpretation, and uniqueness are not enforced.                         |
| §9.4 secure preimage/second-preimage-resistant hash                        | **Fails by default policy**     | Explicitly unfit truncated hashes are accepted.                                                             |
| §9.5 Verifier decides Key Binding policy in advance                        | Conforms at caller-policy level | Caller must opt in; documentation should make downgrade implications explicit.                              |
| §9.7 security-critical claims are not selectively disclosable              | Delegated and undocumented      | Generic disclosure frame permits them. Profile must prohibit them; verifier must validate processed values. |
| §9.7 required validity claims are present                                  | Partial                         | `requiredClaimKeys` exists, but uses ambiguous dot paths.                                                   |
| §9.11 explicit typing                                                      | Delegated                       | Core permits custom `typ`; application profile must define and check it.                                    |
| §9.12 key generation/lifecycle                                             | Out of scope                    | Belongs to injected cryptography and deployment.                                                            |
| §10.2 storage minimization/encryption                                      | Out of scope                    | Belongs to Holder/Issuer/Verifier applications.                                                             |
| §10.3 confidential transport                                               | Out of scope                    | Belongs to protocol integration.                                                                            |
| §10.4 decoy recommendations                                                | Supported                       | Feature exists; policy remains application-specific.                                                        |

## Recommended 1.0 release gate

The following should be treated as release blockers:

1. Implement a strict RFC 9901 parser and a single Section 7.1 processor with runtime schema checks.
2. Reject bare JWT compatibility by default, `alg: none`, malformed/nested reserved fields, wrong Disclosure arity/types, extra array-placeholder members, duplicate Disclosure copies, and all same-level claim collisions.
3. Validate `exp`, `nbf`, `aud`, required claims, and profile-specific claims **after** Disclosure processing; fix `exp === now` and invalid NumericDate handling.
4. Add explicit Issuer audience and KB audience options, KB maximum age/time-window validation, strict KB claim types, and a signature/hash allowlist.
5. Restrict default hash algorithms to security-suitable choices; do not equate IANA registration with suitability.
6. Provide or enforce the Holder reception workflow: verify issuance, reject incoming KB-JWT, then allow presentation.
7. Enforce per-credential salt/digest uniqueness and document/provide a cryptographically secure 128-bit generator contract.
8. Enforce the General JSON first-header rule and apply the same claim-validation pipeline to Compact, Flattened, and General serializations.
9. Replace dot-delimited paths before freezing the 1.0 presentation API.
10. Add official RFC vectors and the full negative matrix to CI.

## Suggested verification API shape

The exact public API is a project decision, but the following concepts should be explicit rather than inferred from the presence of values:

```ts
type VerificationPolicy = {
  allowedIssuerAlgorithms: readonly string[];
  allowedDisclosureHashAlgorithms: readonly string[];
  expectedAudience?: string | readonly string[];
  requiredClaims?: readonly ClaimPath[];
  clockSkewSeconds?: number;
  requireKeyBinding: boolean;
  expectedKeyBindingAudience?: string;
  expectedKeyBindingNonce?: string;
  keyBindingMaxAgeSeconds?: number;
};
```

`ClaimPath` should be an array of string/number segments or a standards-based escaped pointer, not an unescaped dot-delimited string.

Separate role-oriented entry points would also reduce misuse:

- `verifyIssuedSdJwtForHolder()` — requires trailing `~`, rejects KB-JWT, verifies Issuer, returns a validated Holder object;
- `presentValidatedSdJwt()` — accepts only that validated Holder object;
- `verifySdJwtPresentation()` — applies Verifier policy, processed-claim validation, and optional/required Key Binding;
- strict Compact, Flattened JSON, and General JSON decoders feeding the same internal validator.

## Proposed project-submission summary

> A requirement-by-requirement audit of `@sd-jwt/core` at commit `c95e1ab` found that the main RFC 9901 mechanisms are implemented and all current tests pass, but strict verifier conformance is incomplete. Twenty-four negative probes demonstrate acceptance of malformed Disclosure/container shapes, disclosed expired claims, incomplete KB audience/freshness checks, `alg: none`, insecure truncated hash algorithms, an incoming KB-JWT in the Holder presentation path, salt reuse, General JSON header violations, and ambiguous presentation paths. The report recommends treating strict parsing, post-processing validity checks, KB policy checks, secure algorithm policy, Holder-role separation, salt uniqueness, JSON serialization parity, and unambiguous claim paths as 1.0 release blockers.

## Limitations

This was a source and executable behavior audit, not a formal proof, third-party penetration test, or cryptographic implementation certification. The signer, verifier, hasher, salt generator, key storage, trust resolution, transport, and application profile are injected or external; their own security must be assessed separately. The audit is pinned to the commit named above, so later changes require at least a focused re-audit and rerun of the negative matrix.

cre8:
Addresses the RFC 9901 conformance audit findings reported in #388.

## Summary
- Add stricter SD-JWT compact parsing and reject bare JWT inputs as SD-JWTs.
- Validate disclosure structures, `_sd`, array placeholders, nested `_sd_alg`, duplicate disclosure digests, reserved disclosed claim names, and same-level disclosed claim collisions.
- Validate JWT claims after disclosure processing, including strict NumericDate handling, `exp === now`, and expected audience checks.
- Reject `alg: none` and add issuer algorithm allowlist support.
- Tighten KB-JWT validation with strict claim types, `sd_hash` only, expected audience, and max-age checks.
- Reject insecure truncated disclosure hash algorithms by default.
- Reject incoming SD-JWT+KB in Holder presentation flows.
- Enforce duplicate salt/digest rejection during issuance.
- Enforce the General JSON first-unprotected-header rule for `disclosures` and `kb_jwt`.
- Escape presentation paths with literal dots in claim names to avoid ambiguous disclosure selection.
- Update package metadata from draft wording to RFC 9901.

## Validation
- `pnpm exec biome check packages/core/package.json packages/core/src/decode/decode.ts packages/core/src/generalJSON.ts packages/core/src/index.ts packages/core/src/jwt.ts packages/core/src/kbjwt.ts packages/core/src/present/present.ts packages/core/src/sdjwt.ts packages/core/src/types/type.ts packages/core/src/utils/disclosure.ts packages/core/src/test/decode/decode.spec.ts packages/core/src/test/index.spec.ts packages/core/src/test/kbjwt.spec.ts packages/core/test/rfc9901-audit-fixes.spec.ts`
- `pnpm run types:check:tests`
- `pnpm --filter @sd-jwt/core test`

Closes #388


lukasjhan:
@cre8 Wow Mirko, This is super awesome!
Thank you for auditing :)

